PR #1, opened 2017-11-20
Add ExternalMessages.properties to the com/ibm/oti/util package

vsebe:
Copy ExternalMessages.properties to the jdk classes directory.
Include ExternalMessages.properties into resources.jar.

Signed-off-by: Violeta Sebe <vsebe@ca.ibm.com>

### DIFF
--- a/closed/make/CopyFiles.gmk
+++ b/closed/make/CopyFiles.gmk
@@ -1,0 +1,23 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2017 All Rights Reserved
+# ===========================================================================
+# 
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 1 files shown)


jdk/make/CreateJars.gmk
```diff
@@ -271,6 +271,7 @@ $(IMAGES_OUTPUTDIR)/lib$(PROFILE)/_the.rt.jar.contents: $(IMAGES_OUTPUTDIR)/lib$
 	$(MKDIR) -p $(@D)
 	$(RM) $@ $@.tmp
 	$(GREP) -e '\.class$$' $(IMAGES_OUTPUTDIR)/lib$(PROFILE)/_the.jars.contents > $@.tmp
+	$(GREP) -e '\.properties$$' $(IMAGES_OUTPUTDIR)/lib$(PROFILE)/_the.jars.contents >> $@.tmp
```
pshipton: It doesn't seem correct to modify this file which OpenJDK also uses. Somehow the OpenJDK build puts .properties files into the rt.jar without this modification. For example jdk/src/share/classes/javax/sql/rowset/rowset.properties

pshipton:
We discovered all the OpenJDK resources are in resources.jar, OpenJ9 should do the same rather than putting the resource into rt.jar

vsebe:
@pshipton, as discussed, added ExternalMessages.properties to resources.jar instead of rt.jar. Please review, thank you!

PR #3, opened 2017-11-22
fix OpenJ9.gmk

gdams:
currently seeing the following build error:
`OpenJ9.gmk:203: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.`


closed/OpenJ9.gmk
```diff
@@ -200,8 +200,8 @@ stage-j9 : \
 
 # Only update version files when the SHAs change.
 $(OUTPUT_ROOT)/vm/compiler/jit.version : $(call DependOnVariable, OPENJ9_SHA)
-        @$(MKDIR) -p $(@D)
-        $(ECHO) '#define TR_LEVEL_NAME "$(OPENJ9_SHA)"' > $@
+  @$(MKDIR) -p $(@D)
```
DanHeidinga: Should this be a tab?  The version for the old tr.source/jit.version file uses tab

jdekonin: Agreed, it appears to be 2 spaces which is still going to be a problem.

gdams: @DanHeidinga updated PTAL

jdekonin:
Can you please add the 'signed off by' and more detailed comment.

gdams:
@jdekonin updated PTAL

DanHeidinga:
@gdams Thanks for fixing this so quickly!  Sorry I missed the tab vs space when merging the breaking change.

Issue #8, opened 2017-11-30
Java code for dtfj, dtfjview, traceformat is not built


pshipton:
The Java 8 build doesn't build the dtfj, dtfjview or traceformat java code.

While the IBM SDK creates three separate jar files (dtfj.jar, dtfjview.jar, traceformat.jar) and put them into the jre/lib directory, I don't think this is the problem place for them. These are diagnostic tools which don't need to be part of the runtime. I suggest they be put into the tools.jar instead.

The java source code is created by the Java Preprocessor (`$(JPP_JAR)`), using the following preprocessor configs: SIDECAR18-DTFJ, SIDECAR18-DTFJVIEW, SIDECAR18-TRACEFORMAT

groeges:
@pshipton Have added the Java Preprocessor to get missing code but I am having issues getting code to compile (as separate jars at the moment). Some of the code is complaining about some zOS classes/code even though I am running on a xLinux machine. Who is best person to help me resolve this?

groeges:
@pshipton  Looking at 
```
./openj9/jcl/src/openj9.dtfj/share/classes/com/ibm/dtfj/corereaders/ClosingFileReader.java
```
it contains pre-processor lines that say:
```
/*[IF PLATFORM-mz31 | PLATFORM-mz64 | ! Sidecar19-SE]*/
```
but when running the pre-processor we use SIDECAR18-DTFJ, which doesn't set the fag for Sidecar19-SE :-(  and this will then add the zOS code.

Any ideas?

pshipton:
@keithc-ca was just chatting to me about this. I believe we can fix it by using the Sidecar18-SE-OpenJ9 flag.

keithc-ca:
Assuming my test builds encounter no issues, I'll open a PR soon.

pshipton:
Fixed by eclipse/openj9#746

groeges:
@pshipton, I presume this PR only resolved the issue I was seeing with the pre-processor and not the entire issue of the actual generation of the rtf.jar and others.

groeges:
Meant dtfj.jar

pshipton:
Yup, I've updated the comment

groeges:
[...]

PR #9, opened 2017-12-04
Update make file to generate missing executables

groeges:
Executables that are being generated are:
jdmpview, traceformat (for all)
and jextract (for Linux and AIX only) 

Signed-off-by: Steve Groeger GROEGES@uk.ibm.com


groeges:
@andrew-m-leonard Please can you review this PR to add the executables on JDK8.
They will not work currently until issue: https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/8 is resolved.

PR #15, opened 2017-12-11
Changes to allow diagnostic jars to be created

groeges:
The jars that are created are:
dtfj.jar, dtfjview.jar and traceformat.jar

Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -291,6 +291,18 @@ $(J9JCL_SOURCES_DONEFILE) : $(AllJclSource)
 			-dest $(JDK_OUTPUTDIR)/j9jcl_sources/jdk/src/share/classes \
 			-macro:define "com.ibm.oti.vm.library.version=29" \
 			-tag:define "PLATFORM-$(OPENJ9_PLATFORM_CODE)"
+	@$(BOOT_JDK)/bin/java \
+		-cp $(OUTPUT_ROOT)/vm/sourcetools/lib/jpp.jar \
+		-Dfile.encoding=US-ASCII \
+		com.ibm.jpp.commandline.CommandlineBuilder \
+			-verdict \
+			-baseDir $(OPENJ9_TOPDIR)/ \
+			-config SIDECAR18-TOOLS-OPENJ9 \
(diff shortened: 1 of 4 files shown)


groeges:
@andrew-m-leonard Please can you review and merge this PR.

pshipton:
#8 I specifically said these jar files should not go into the lib/ext directory, but the packages / classes should go into the tools.jar instead.

groeges:
@pshipton Well to be exact you said "I suggest they be put into the tools.jar instead" !

In the IBM JDK 8 they are in those jars in the lib/ext directory. Are we OK moving these to tools.jar and not having the dtfj, dtfjview and traceformat jars. If some was using those jars in IBM JDK 8 and move to OpenJ9 with OpenJDK8, they will not have these jars and will need to modify their code  etc to use the new tools.jar.   If we deem this OK then I can easily move these classes to tools.jar.

pshipton:
OpenJ9 is not the same as IBM Java, and doesn't need to make the same mistakes. We also don't expect IBM customers to switch to OpenJDK. If I wasn't ok with moving the code, I wouldn't have "suggested" it, but I'm open to other opinions.

Putting these diagnostic/development tools into the ext directory incurs a small runtime performance penalty every time a non-bootstrap class is loaded (using the default app classloader). Since the jars are on the extension classloader search path they are searched first before any application class is loaded by the default app classloader. As these are diagnostic/development tools which aren't needed by the runtime, in Java 8 they belong in the tools.jar along with other similar tools. The tools are mostly used via the executables (jmpview, traceformat, jextract) and as long as these tools work there shouldn't be any problem. There are 2 caveats. One is dtfj which is an API you can write code against. This will be a change in behavior as dtfj apps will need to put tools.jar on the path to work. Since dtfj is a diagnostic tool I think this is the better option to affecting runtime performance. The second is traceformat. In the past there wasn't an executable for this and the IBM docs specify how to run it using a main class. Again the tools.jar will be required to run it, but since the executable is much easier to use I don't see a big issue.

groeges:
@pshipton That is OK. I have moved them to tools.jar and will submit a new pull request.

PR #16, opened 2017-12-11
Put dtfj, dtfjview and traceformat classes in tools.jar

groeges:
Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

### DIFF
--- a/jdk/make/CreateJars.gmk
+++ b/jdk/make/CreateJars.gmk
@@ -56,47 +56,6 @@ $(eval $(call MakeDir, $(IMAGES_OUTPUTDIR)/lib))
 
 ##########################################################################################
 
-$(eval $(call SetupArchive,BUILD_DTFJ_JAR, , \
-    SRCS := $(JDK_OUTPUTDIR)/classes, \
-    INCLUDES := com/ibm/dtfj com/ibm/jvm/j9/dump com/ibm/java/diagnostics/utils, \
-    JAR := $(IMAGES_OUTPUTDIR)/lib/ext/dtfj.jar, \
-    SKIP_METAINF := true))
-
-
-##########################################################################################
-
-$(eval $(call SetupArchive,BUILD_DTFJVIEW_JAR, , \
-    SRCS := $(JDK_OUTPUTDIR)/classes, \
-    INCLUDES := com/ibm/jvm/dtfjview, \
-    JAR := $(IMAGES_OUTPUTDIR)/lib/ext/dtfjview.jar, \
-    SKIP_METAINF := true))
-
-
-##########################################################################################
-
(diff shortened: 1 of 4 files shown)


groeges:
@andrew-m-leonard As per @pshipton I have moved the classes for these jars to the tools.jar instead of having them as separate jars in the lib/ext directory. 
This PR does just that.  Please can you review and merge if OK.

jdk/make/CreateJars.gmk
```diff
@@ -590,7 +555,20 @@ $(eval $(call SetupArchive,BUILD_TOOLS_JAR, , \
         META-INF/services/com.sun.jdi.connect.spi.TransportService \
         META-INF/services/com.sun.tools.attach.spi.AttachProvider \
         META-INF/services/com.sun.tools.internal.ws.wscompile.Plugin \
-        META-INF/services/com.sun.tools.internal.xjc.Plugin, \
+        META-INF/services/com.sun.tools.internal.xjc.Plugin \
+        com/ibm/jvm/Debug.class \
```
pshipton: Where do these classes come from? I don't find them.

pshipton:
Do the executables still work with these changes? I expected a change so they look in the tools.jar, which isn't on the classpath by default.

jdk/make/CreateJars.gmk
groeges: Which classes are you referring to? If you mean the META-INF/services/.... ones, then these were in that make file already and I presumed they are related to OpenJDK8. If you mean the com/ibm/jvm/... ones then these were listed in the icl-build.mk file that I was pointed to which was previously used to build the tracefomat.jar. These classes do not exist as separate java files (which caused me some issues initially) but they are classes that are defined in other java classes (which I find very strange!!!).

groeges:
In my build they still seem to work. I run `./build/linux-x86_64-normal-server-release/images/j2sdk-image/bin/traceformat` and that works.
Hmm, if I run `./build/linux-x86_64-normal-server-release/images/j2sdk-image/jre/bin/traceformat` it doesn't as it can't find the class, as you expected. 
Will try and see if I can get these working!!!

pshipton:
It is expected that traceformat doesn't work in the jre image, as the jre image doesn't contain tools.jar. The executables should only be present in the bin directory and only part of the sdk image. They should not be present in the jre/bin directory and part of the jre image. This is similar to other tools like javac, rmic, etc.

jdk/make/CreateJars.gmk
pshipton: Ok I see, they are defined in the com.ibm.jvm.TraceFormat.java file.

Issue #23, opened 2018-01-10
AO8/J9 s390x builds do not work out of the box


sxa555:
I presume this is somewhere around the `redirector.c` code within `openj9`

if you download the s390x AO8/J9 builds from www.AdoptOpenJDK.net and try to run it you get this error:

```
jcktest@csz25074:/dev/shm/jdk8u152-b16/bin$ ./java -version
Error: missing `server' JVM at `/dev/shm/jdk8u152-b16/jre/lib/s390x/server/libjvm.so'.
Please install or use the JRE or JDK that contains these missing components.
jcktest@csz25074:/dev/shm/jdk8u152-b16/bin$ 
```
Creating a symbolic link from `jre/lib/s390x/server` to `jre/lib/s390x/j9vm` allows the JVM to start up, so somewhere along the lines in the build it's still looking for `libjvm.so` in HotSpot's location.

pshipton:
Its caused by the jvm.cfg file, which needs to contain

```
-j9vm KNOWN
-server IGNORE
```

sxa555:
The one in `jdk/src/solaris/bin/s390x` looks correct, but that isn't the one that's in the final distribution

sxa555:
I think this is because default (non-OpenJ9) build on `s390x` is `zero` not `server`, which means `JVMCFG_ARCH` in `CopyFiles.gmk` ends up as `zero` instead of `s390x` so it doesn't copy the correct file. @pshipton I'm not sure where the fix for this needs to be. I'm happy to do it, but want it in the "correct" repo wherever that is.

pshipton:
Strangely when we do builds they don't have the same issue, it picks up a jvm.cfg with `-j9vm KNOWN` and the build runs. We don't seem to be doing any special i.e. `bash ./configure --with-boot-jdk=/sdk/openjdk7/linux_390-64_cmprssptrs --enable-cuda --with-freemarker-jar=/tools/freemarker-2.3.8/lib/freemarker.jar && make images`. 

We don't want to have modifications to OpenJDK so modifying `jdk/make/CopyFiles.gmk` is not ideal, unless the modifications can be contributed to OpenJDK.

sxa555:
[...]

PR #24, opened 2018-01-10
Fix to pick up the correct libjvm.so on s390x OpenJ9 builds

sxa555:
Fixes https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/23

It may also be sensible to remove the `s390x/jvm.cfg`if it's not being picked up, but I understand it does work on some people's machines.

### DIFF
--- a/jdk/src/solaris/bin/zero/jvm.cfg
+++ b/jdk/src/solaris/bin/zero/jvm.cfg
@@ -1,4 +1,7 @@
-# Copyright (c) 2009, 2013, Oracle and/or its affiliates. All rights reserved.
+#
+# (c) Copyright IBM Corp. 2003, 2018 All Rights Reserved
+#
+# Copyright (c) 2003, 2013, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -31,5 +34,5 @@
 # "-XXaltjvm=<jvm_dir>" option, but that too is unsupported
 # and may not be available in a future release.
 #
--server KNOWN
--client IGNORE
+-j9vm KNOWN
+-server IGNORE


jdk/src/solaris/bin/zero/jvm.cfg
```diff
@@ -31,5 +34,8 @@
 # "-XXaltjvm=<jvm_dir>" option, but that too is unsupported
 # and may not be available in a future release.
 #
--server KNOWN
--client IGNORE
+-j9vm KNOWN
+-hotspot IGNORE
```
pshipton: Although most of the jvm.cfg files contain them, these lines for `-hotspot, -classic, -native, -green` seem to have been copied from some really old build. When we updated the jvm.cfg for Windows, I suggested minimal changes, i.e. set `-server IGNORE` and add `-j9vm KNOWN`.

jdk/src/solaris/bin/zero/jvm.cfg
```diff
@@ -1,4 +1,7 @@
-# Copyright (c) 2009, 2013, Oracle and/or its affiliates. All rights reserved.
+#
+# (c) Copyright IBM Corp. 2003, 2017 All Rights Reserved
```
pshipton: Please set 2018

jdk/src/solaris/bin/zero/jvm.cfg
sxa555: I mostly left them in because they were there on the `s390x/jvm.cfg` that was already in the repository. I'll remove them

jdk/src/solaris/bin/zero/jvm.cfg
sxa555: Fixed - Good spot - I read 2017 and thought "Yeah that's up to date" ... #JanuaryFever

sxa555: Done

jdk/src/solaris/bin/zero/jvm.cfg
```diff
@@ -31,5 +34,5 @@
 # "-XXaltjvm=<jvm_dir>" option, but that too is unsupported
 # and may not be available in a future release.
 #
--server KNOWN
--client IGNORE
+-j9vm KNOWN
+-server IGNORE
```
pshipton: You can keep the `-client IGNORE` line.

keithc-ca:
It seems that `JVM_VARIANT_ZERO` must be `true`: why would we want that?

CopyFiles.gmk line 274:
```
# How to install jvm.cfg.
#
ifeq ($(JVM_VARIANT_ZERO), true)
  JVMCFG_ARCH := zero
else
  JVMCFG_ARCH := $(OPENJDK_TARGET_CPU_LEGACY)
endif
```

DanHeidinga:
@sxa555 Given the Adopt change: https://github.com/AdoptOpenJDK/openjdk-build/pull/203 is this still required?  I'm guessing not.

sxa555:
@DanHeidinga With the build now complete and verification that it has solved the problem, no it's not required, therefore closing :-)
Indeed @keithc-ca it's a logic mistake in the build script which has now been resolved in https://github.com/AdoptOpenJDK/openjdk-build/commit/99c2bbdc2dad12924d41ba0526c4fec72bdd814f

PR #41, opened 2018-02-12
Fix git queries for OPENJDK_TAG

keithc-ca:
Only match openjdk tags whose names start with 'jdk8u'.

Adjust whitespace as requested for #35.

### DIFF
--- a/common/autoconf/configure
+++ b/common/autoconf/configure
@@ -122,16 +122,17 @@ fi
 ### Process command-line arguments
 ###
 # Add default OpenJ9 closed extensions
-full_script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-OPENJDK_TAG=`git -C $full_script_dir/../.. describe --abbrev=0 --tags`
+full_src_root_dir="$( cd "${conf_script_dir}/../.." && pwd )"
+OPENJDK_SHA=`git -C $full_src_root_dir rev-parse --short HEAD`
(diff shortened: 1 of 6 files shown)


pshipton:
jdk/make/closed/autoconf/custom-hook.m4 also needs to be fixed the same way.

The description of the commit could be more explicit, i.e. "Fix openjdk tag query" or "Fix OPENJDK_TAG query"

common/autoconf/configure
```diff
@@ -123,10 +123,10 @@ fi
 ###
 # Add default OpenJ9 closed extensions
 full_script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-OPENJDK_TAG=`git -C $full_script_dir/../.. describe --abbrev=0 --tags`
+OPENJDK_TAG=`git -C $full_script_dir/../.. describe --abbrev=0 --tags --match "jdk8u*"`
```
pshipton: JDK9 contains `--match "jdk-9*" "${OPENJDK_SHA}"`. Does the SHA also need to be specified here?

keithc-ca: I don't think it is needed; git assumes 'HEAD', but I'll add it for consistency.

keithc-ca:
custom-hook.m4 (and files derived via autogen.sh) updated along with commit description.

pshipton:
This repo doesn't support pull request testing yet.

PR #52, opened 2018-02-28
Rework rules to keep version files up-to-date

keithc-ca:
The function DependOnVariable does not exist for Java 8, thus its use
had no effect. To ensure files have the desired content we capture it
in a temporary file and only update the target file if it is absent or
has the wrong content.

Signed-off-by: Keith W. Campbell <keithc@ca.ibm.com>


keithc-ca:
See also https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/135.

pshipton:
jenkins compile plinux

PR #59, opened 2018-03-08
Enable DDR support for Linux x86-64

keithc-ca:
* add configure option '--enable-ddr' (enabled by default for xa64)
* enable opt_useOmrDdr accordingly
* build j9ddr.jar for inclusion in JDK

Depends on https://github.com/eclipse/openj9/pull/1399.

### DIFF
--- a/closed/make/DDR-jar.gmk
+++ b/closed/make/DDR-jar.gmk
@@ -1,0 +1,59 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.
+#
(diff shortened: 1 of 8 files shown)


keithc-ca:
The copyright failure is with a generated file: common/autoconf/generated-configure.sh.

closed/make/DDR-jar.gmk
```diff
@@ -0,0 +1,59 @@
...
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+# ===========================================================================
+
+# The main target.
+all :
+	@$(ECHO) Finished building j9ddr.jar
```
pshipton: JDK9 doesn't have this.

keithc-ca: Agreed it's not necessary: removed by https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/61.

PR #62, opened 2018-03-12
Fix race condition in jdwp

pdbain-ibm:
Normally, the event helper thread suspends all threads, then the debug loop in
the listener thread receives a command to resume.  The debugger may deadlock if
the debug loop in the listener thread starts processing commands (e.g. resume
threads) before the event helper completes the initialization (and suspends
threads).

This patch adds synchronization to ensure the event helper completes the
initialization sequence before debugger commands are processed.

@andrew-m-leonard FYI.

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>

### DIFF
--- a/jdk/src/share/back/debugInit.c
+++ b/jdk/src/share/back/debugInit.c
@@ -1,3 +1,24 @@
+/*******************************************************************************
+ * (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+ *
+ * This program and the accompanying materials are made available under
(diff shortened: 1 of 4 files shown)


pdbain-ibm:
@andrew-m-leonard @adamfarley would you kindly review this?
Thanks
     -p

pshipton:
Similar problems as https://github.com/ibmruntimes/openj9-openjdk-jdk9/pull/138. @pdbain-ibm

PR #65, opened 2018-03-15
Restore creation of dtfj, dtfjview & traceformat jars

keithc-ca:
* Enables basic commands in jdmpview.
* Content in those jars forms part of the API to an SDK. Used, for example, by HealthCenter.
* Include only necessary OpenJ9 classes in tools.jar.
* Revert many changes to jdk/make/CreateJars.gmk by moving as much as possible to closed/make/CreateJars.gmk.
* Use default APP_CLASSPATH for jdmpview, jextract and traceformat.
* Add missing IBM copyright notice.

### DIFF
--- a/closed/make/CreateJars.gmk
+++ b/closed/make/CreateJars.gmk
@@ -1,0 +1,107 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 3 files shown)


jdekonin:
jenkins plinux

pshipton:
I'm not convinced this is the correct way to go. Although having the jars in the ext directory is how it works for IBM Java 8, its not the correct way to do things. Putting these jars in the ext directory incurs some costs. Both in the disk size of a JRE (which shouldn't contain these development tools), and also in class loading performance at runtime. Any class lookup based on the system classloader will be a little slower because of these additional jar files are in the ext directory.

DanHeidinga:
@pshipton Do you have an alternative proposal for how to handle this?  If you do, can you create an issue to track the discussion?

pshipton:
@DanHeidinga I think it was correct the way it was originally before this PR was merged. I suppose we shouldn't just revert this PR since moving changes to closed/make is desired. I've opened #67 and https://github.com/eclipse/openj9/issues/1539.

PR #71, opened 2018-04-09
Improve detection of latest JDK tag

keithc-ca:
Now detects both annotated and unannotated tags.


andrew-m-leonard:
jenkins compile plinux

PR #72, opened 2018-04-13
Change compilation setup for DDR classes

keithc-ca:
Previously, the compile step specified `GENERATE_USINGJDKBYTECODE` which uses `JVM := $(JAVA_SMALL)` with a 512M heap. This lead to OutOfMemoryError on Windows.

With this change, the compile step specifies `GENERATE_JDKBYTECODE` which uses `JVM := $(JAVA)` with a 1100M or 1600M heap depending on whether the boot JDK is 32-bits or 64-bits.


pshipton:
jenkins test sanity plinux

Issue #79, opened 2018-05-03
openj9-notices.html is out of date


pshipton:
The content of https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/openj9-notices.html is out of date wrt OpenJ9 https://github.com/eclipse/openj9/blob/master/longabout.html

The file is copied into the jdk and jre images when builds are created. Rather than update the copy, a better solution is to delete openj9-notices.html, and copy the OpenJ9 file to the images instead, which can be renamed to openj9-notices.html during the copy.

pshipton:
Note this affects all versions of the extensions.

adamfarley:
"Out of date" - I see that the date at the top of the jdk8 notices file is a month later than the date at the top of the openj9 notices file. When we consider the bits in the jdk8 file which are not in the openj9 file (like the bit near the top about Third Party Content, and the bit about libauxv), it looks like a merge would be more appropriate here. 

However, Peter initial comment appears to explicitly rule this out as an option, so perhaps he knows something I don't.

@pshipton - Thoughts?

pshipton:
The idea is to use the original OpenJ9 file, rather than maintain a copy of it in all the extensions repos. A copy in the extensions repos is not required, the OpenJ9 file can be copied to the correct place in the images. This avoids having to update the copy any time the original file is updated. Already the copies were not updated in a timely fashion when the original changed, and it was only discovered by accident they were out of sync. It may be a little more work initially to use the original file instead of the copy, but in the longer term I believe its worth the effort.

adamfarley:
Then do you advise we move the musl bit into the OpenJ9 file, or do without it when we copy the openJ9 file instead of the extensions notices file? I don't see a third option, unless we want to copy both files, or perform some kind of merge at build time, both of which sound fraught with complications.

pshipton:
I don't understand the question. The OpenJ9 file https://github.com/eclipse/openj9/blob/master/longabout.html contains the musl license. This file just needs to be copied into the images (see references to openj9-notices.html in OpenJ9.gmk) and renamed to openj9-notices.html and then https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/openj9-notices.html can be deleted.

pshipton:
libauxv is not included in OpenJ9 and was removed from the OpenJ9 longabout.html. It is used by OMR however the license was changes, see https://github.com/eclipse/omr/pull/2171

[...]

PR #85, opened 2018-05-15
Use the openj9 notices file for builds instead of the extensions one

adamfarley:
When building Java, we copy the extensions notices file into the
build directory. Since every extensions repository has its own copy,
this means that every update needs to be copied into every extensions
repository, making upkeep difficult.

If, however, we use the copy in the openj9 repository, maintenance
becomes easier because every version of OpenJ9 Java uses the same
Openj9 VM, so we only have to modify the one notices file during
maintenance.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -73,7 +73,8 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
   OPENJ9_VM_FILES += j9ddr.dat
 endif
 
-OPENJ9_NOTICE_FILES := openj9-notices.html
+OPENJ9_NOTICE_FILE := openj9/longabout.html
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Jenkins compile pLinux

andrew-m-leonard:
Compile pLinux

pshipton:
The missing part of this is to delete https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/openj9-notices.html

andrew-m-leonard:
oh yeah, doh!

adamfarley:
Thought about that and decided removing a notices file from a repository was a different issue, with its own attendant legal questions. Continuing this discussion in the issue (https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/79)

PR #86, opened 2018-05-16
Remove superfluous notices file

adamfarley:
We used to copy this file into Java when we build it. Now,
however, we use the copy in the openj9 repository, so the extensions
copy is no longer required.


pshipton:
jenkins compile plinux

PR #95, opened 2018-06-15
Copy configured values to relevant UMA properties in .spec files

keithc-ca:
Fixes Java 8 portion of eclipse/openj9#1919.


pshipton:
jenkins compile all

PR #97, opened 2018-06-16
Update IBM version to 8.0.0.2 for jdk8u172-b11 and OpenJ9 0.9.0

pshipton:


andrew-m-leonard:
This needs doing on the new openj9-0.9 banch

pshipton:
I thought we should make this change before the branch was created, but too late now.
Replaced by #98

PR #100, opened 2018-06-28
Enable DDR by default for Windows 32-bit builds

keithc-ca:
Depends on eclipse/openj9#2266


pshipton:
jenkins test all win32

PR #102, opened 2018-07-09
Prevent the ClassLoader object being GCd in the middle of a method

adamfarley:
In ResourceBundle, the private method "getBundleImpl" takes
a ClassLoader parameter, but doesn't appear to use it directly.
If a GC cycle occurs during the method execution, we can lose that
ClassLoader, causing a bug whose most visible symptom is us being
unable to find a locale-specific entity that definitely exists.

E.g. If we have Stuff_fr.class and Stuff_fr_CA.class, and we ask for
"Stuff" with the locale "fr, CA", then a garbage collection here
may cause us to return Stuff_fr.class (as if Stuff_fr_CA.class did not
exist).

Using the loader parameter for an operation at the very end of the
method prevents GC from collecting it, and prevents this problem.

### DIFF
--- a/jdk/src/share/classes/java/util/ResourceBundle.java
+++ b/jdk/src/share/classes/java/util/ResourceBundle.java
@@ -25,7 +25,7 @@
 
 /*
  * (C) Copyright Taligent, Inc. 1996, 1997 - All Rights Reserved
- * (C) Copyright IBM Corp. 1996 - 1999 - All Rights Reserved
+ * (C) Copyright IBM Corp. 1996 - 2018 - All Rights Reserved
  *
  * The original version of this source code and documentation
  * is copyrighted and owned by Taligent, Inc., a wholly-owned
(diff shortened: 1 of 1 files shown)


jdk/src/share/classes/java/util/ResourceBundle.java
```diff
@@ -1398,7 +1398,13 @@ private static ResourceBundle getBundleImpl(String baseName, Locale locale,
             bundle = baseBundle;
         }
 
-        return bundle;
+        //The OpenJ9 GC might collect the loader before we return here. This prevents that.
+        try {
+            return bundle;
+        } finally {
+            //Should never be true. Using the loader here prevents it being GC'd.
+            if (baseName.equals(loader)) throw new Error("Unexpected error.");
```
pshipton: This could use == instead of .equals(), which should be a faster test.

adamfarley: Incomparable object. Will use a global Object for the compare.

pshipton:
jenkins test sanity plinux

pshipton:
Note this change should also be applied to the openj9-0.9 branch.

adamfarley:
@pshipton - Good catch. Here you go.

https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/103

PR #105, opened 2018-07-19
Enable CMake build

dnakamura:
Signed-off-by: Devin Nakamura <devinn@ca.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -314,6 +314,7 @@ run-preprocessors-j9 : stage-j9
 		&& cd $(OUTPUT_ROOT)/vm \
 		&& $(MAKE) $(MAKEFLAGS) -f buildtools.mk \
 			BUILD_ID=$(BUILD_ID) \
+			CMAKE=$(CMAKE) \
 			FREEMARKER_JAR="$(FREEMARKER_JAR)" \
 			J9VM_SHA=$(OPENJ9_SHA) \
 			JAVA_HOME=$(BOOT_JDK) \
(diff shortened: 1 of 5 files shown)


closed/OpenJ9.gmk
```diff
@@ -324,14 +325,22 @@ run-preprocessors-j9 : stage-j9
 			UMA_OPTIONS_EXTRA="-buildDate $(shell date +'%Y%m%d')" \
 			tools \
 	)
+ifneq (true,$(OPENJ9_ENABLE_CMAKE))
 	@$(SED) -i -e 's/^\s*VERSION_MAJOR\s*.*=.*/VERSION_MAJOR:=8/' $(OUTPUT_ROOT)/vm/makelib/mkconstants.mk
```
pshipton: I can't see that this is still required, considering mkconstants.mk doesn't assign VERSION_MAJOR to anything any more. I think this can just be deleted.

closed/OpenJ9.gmk
```diff
@@ -324,14 +325,22 @@ run-preprocessors-j9 : stage-j9
...
+endif
 
 build-j9 : run-preprocessors-j9
 	@$(ECHO) Compiling OpenJ9 in $(OUTPUT_ROOT)/vm
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+	(export OPENJ9_BUILD=true $(EXPORT_MSVS_ENV_VARS) \
+		&& $(MAKE) -C  $(OUTPUT_ROOT)/vm/build $(MAKEFLAGS) install \
```
pshipton: How does cmake set VERSION_MAJOR=8 ?

dnakamura: It gets set further up,  when we call `buildtools.mk`, which will [pass it on to cmake](https://github.com/eclipse/openj9/blob/82661a78819cb7de3b80fabd69158d396a62f980/runtime/buildtools.mk#L53)

closed/OpenJ9.gmk
pshipton: @dnakamura if you remove this sed command completely I will PR test this and then merge if all goes well.

pshipton:
jenkins compile xlinux

PR #106, opened 2018-07-27
Fixes bug dbgsysBuildLibName expects single path

theresa-m:
Fixes: https://github.com/eclipse/openj9/issues/2129

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

@pshipton

### DIFF
--- a/jdk/src/windows/back/linker_md.c
+++ b/jdk/src/windows/back/linker_md.c
@@ -1,4 +1,10 @@
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * Copyright (c) 1998, 2013, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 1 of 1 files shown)


jdk/src/windows/back/linker_md.c
```diff
@@ -100,19 +109,17 @@ dbgsysGetLastErrorString(char *buf, int len)
...
     const int pnamelen = pname ? (int)strlen(pname) : 0;
 
     *holder = '\0';
-    /* Quietly truncates on buffer overflow. Should be an error. */
-    if (pnamelen + (int)strlen(fname) + 10 > holderlen) {
-        return;
-    }
 
-    if (pnamelen == 0) {
+    if ((pnamelen == 0) && ((int)strlen(fname) + 10 <= holderlen)) {
```
pshipton: not sure we need the +10, just +5 for ".dll" and the null terminator.

jdk/src/windows/back/linker_md.c
```diff
@@ -51,11 +57,14 @@ static void dll_build_name(char* buffer, size_t buflen,
     path = strtok_s(paths_copy, PATH_SEPARATOR, &next_token);
 
     while (path != NULL) {
-        _snprintf(buffer, buflen, "%s\\%s.dll", path, fname);
-        if (_access(buffer, 0) == 0) {
-            break;
+        /* If the path will cause a buffer overflow skip it. This path is invalid. */
+        if (path + (int)strlen(fname) + 10 <= buflen) {
```
pshipton: This needs to use the length of the path. 

Could pull strlen(fname) out of the loop.

Not sure of the purpose of +10. I see +6 for the extra chars added plus the null terminator.

keithc-ca: `strlen` returns `size_t`: the cast to `int` is inappropriate and should not be necessary.

pshipton:
obsoleted by #107

PR #107, opened 2018-08-02
Fix handling of path list in dbgsysBuildLibName

keithc-ca:
Silently ignore paths that don't fit in the supplied buffer.

This should address https://github.com/eclipse/openj9/issues/2129.


pshipton:
jenkins test sanity win

pshipton:
jenkins test sanity win32

PR #114, opened 2018-09-14
Enable building a 'test-image'

keithc-ca:
Java 11 and newer build frameworks allow building a 'test-image' which include test artifacts (executables and shared libraries). This does the same for Java 8.

Add make target 'test-image'
* add 'test-image' as a prerequisite for 'images'.

Remove unused rules
* neither 'product-images' nor 'ALL_TARGETS' is mentioned in other makefiles, so they can safely be removed along with their prerequisites.

See https://github.com/eclipse/openj9/issues/2100.


pshipton:
jenkins test sanity zlinux jdk11

pshipton:
jenkins test sanity zlinux

PR #117, opened 2018-10-03
Compile DDR_VM code twice

keithc-ca:
* once with generated source files
* again with generated class files to validate blob integrity
* stop staging debugtools
* merge DDR support makefiles

Once all Java versions validate their blobs this way, a change can be introduced in openj9 that distinguishes core files created after a certain time allowing pointer classes to be derived at runtime, fixing eclipse/openj9#2507.


pshipton:
@keithc-ca please take a look at the build failures

keithc-ca:
Updated for https://github.com/eclipse/openj9/pull/3190.

keithc-ca:
jenkins test all plinux

keithc-ca:
The failures are because the boot JDK is Java 7. I'll either need to use the JVM being built or find another source for ASM.

pshipton:
jenkins test sanity win32 jdk8

PR #118, opened 2018-10-03
Initial macosx support in the configuration files

DanHeidinga:
Regenerate with:
  bash common/autoconf/autogen.sh

issue: eclipse/openj9#36

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

### DIFF
--- a/jdk/make/closed/autoconf/custom-hook.m4
+++ b/jdk/make/closed/autoconf/custom-hook.m4
@@ -193,6 +193,13 @@ AC_DEFUN_ONCE([OPENJ9_PLATFORM_SETUP],
       else
         OPENJ9_BUILDSPEC="win_x86-64_cmprssptrs"
       fi
+    elif test "x$OPENJDK_BUILD_OS" = xmacosx; then
+      OPENJ9_PLATFORM_CODE=oa64
+      if test "x$OPENJ9_LIBS_SUBDIR" = xdefault; then
+        OPENJ9_BUILDSPEC="osx_x86-64"
+      else 
+        OPENJ9_BUILDSPEC="osx_x86-64_cmprssptrs"
+      fi
     else
       AC_MSG_ERROR([Unsupported OpenJ9 platform ${OPENJDK_BUILD_OS}!])
     fi
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -1004,6 +1004,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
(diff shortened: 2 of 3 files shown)


DanHeidinga:
FYI @keithc-ca

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -193,6 +193,9 @@ AC_DEFUN_ONCE([OPENJ9_PLATFORM_SETUP],
       else
         OPENJ9_BUILDSPEC="win_x86-64_cmprssptrs"
       fi
+    elif test "x$OPENJDK_BUILD_OS" = xmacosx; then
+      OPENJ9_PLATFORM_CODE=os64
```
keithc-ca: This will require adding `PLATFORM-os64` to the list of `platforms` in `jpp_configuration.xml` in openj9 can you open a PR for that first?

babsingh: in **openj9-openjdk-jdk9** extensions repo, `oa64` is used instead of `os64`: [code_link](https://github.com/ibmruntimes/openj9-openjdk-jdk9/blob/openj9/closed/autoconf/custom-hook.m4#L221).

also, we should add support for `cmprssptrs` here:
```
    ......
    elif test "x$OPENJDK_BUILD_OS" = xmacosx; then
      OPENJ9_PLATFORM_CODE=oa64
      if test "x$OPENJ9_LIBS_SUBDIR" = xdefault; then
        OPENJ9_BUILDSPEC="osx_x86-64"
      else
        OPENJ9_BUILDSPEC="osx_x86-64_cmprssptrs"
      fi
    ......
```

DanHeidinga: Good catch on `oa` vs `os`.  The os is a typo

keithc-ca:
eclipse/openj9#3127 is merged; after fixing the typo, this should be good to go.

DanHeidinga:
> (without which, osx could not build);

@keithc-ca this can't possibly be true as we've built osx has been built by multiple devs without this change.

>  after fixing the typo, this should be good to go.

Typo is fixed

keithc-ca:
> this can't possibly be true

My mistake; the JCL preprocessor will let you define whatever tags you want, but you couldn't *use* `PLATFORM-oa64` in source code without the jpp config change.

keithc-ca:
Ignoring expected copyright check failure for `common/autoconf/generated-configure.sh`.

Issue #119, opened 2018-10-04
Add openssl to the third party licenses file


lumpfish:
As a result of bundling openssl libraries with the jdk the openssl license needs to be added to the third party license notices file included in the built jdk.

mstoodle:
I'm not sure I understand this one.... openj9-openjdk-jdk8 is not bundling an openssl library with its source code, so why is this license needed?

It's AdoptOpenJDK that will be doing the bundling, by specifying the appropriate configure options. How is this different than the font libraries or certificates that are also built into the OpenJDK binaries at AdoptOpenJDK? Those things aren't included in OpenJDK's license files are they?

I suppose we could provide it for "informational purposes" *in case* someone chooses to bundle the library.

andrew-m-leonard:
@mstoodle true... I guess if someone builds their own binary using the --with-openssl... then they ought to be aware...

andrew-m-leonard:
at the moment I am thinking of copying https://www.openssl.org/source/license.html to a new extensions licenses file, which then gets added to the built jdk...

mstoodle:
ok, that doesn't sound unreasonable to me

lumpfish:
Fixed by https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/121

PR #121, opened 2018-10-05
Add OpenSSL Toolkit license

andrew-m-leonard:


andrew-m-leonard:
jenkins compile xlinux

closed/OpenJ9.gmk
```diff
@@ -88,6 +88,8 @@ OPENJ9_NOTICE_FILE := openj9/longabout.html
 OPENJ9_NOTICE_FILE_RENAME := openj9-notices.html
 OPENJ9_REDIRECTOR := redirector/$(LIBRARY_PREFIX)jvm$(SHARED_LIBRARY_SUFFIX)
 
+OPENSSL_NOTICE_FILE := openssl-license
```
lumpfish: openjdk and openj9 just have asingle file containing all the third party notices.  Calling the file openssl-license implies more files if more packages were to be added in future.  So why not have a single file (openj9-openjdk-notices.html?) and put the license and preamble in there, then if there are future packages the build won't need to be changed again?

openssl-license
```diff
@@ -0,0 +1,131 @@
+This notice is provided with respect to the OpenSSL toolkit, which may be
+included with the JRE and JDK.
+
```
lumpfish: I think we should add words to the effect that if the user does not want to install openssl they can safely delete it from the jdk.

lumpfish:
Fixes https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/119

Issue #123, opened 2018-10-05
openssl doesn't build on aix in OpenJ9 builds


pshipton:
Re #112

On aix using
get_source.sh --openssl-version=1.1.1
configure with --with-openssl=fetched --enable-openssl-bundling

```
15:36:02 ar: creating an archive file apps/libapps_a.a
15:36:19      147  1500-010: (W) WARNING in async_start_func: Infinite loop.  Program may not stop.
15:37:03 "crypto/ec/ecp_nistz256.c", line 227.31: 1506-1298 (W) The subscript 4 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 228.31: 1506-1298 (W) The subscript 5 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 229.31: 1506-1298 (W) The subscript 6 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1251.37: 1506-1298 (W) The subscript 4 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1251.48: 1506-1298 (W) The subscript 5 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1251.59: 1506-1298 (W) The subscript 6 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1251.70: 1506-1298 (W) The subscript 7 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1252.37: 1506-1298 (W) The subscript 4 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1252.48: 1506-1298 (W) The subscript 5 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1252.59: 1506-1298 (W) The subscript 6 is out of range. The valid range is 0 to 3.
15:37:03 "crypto/ec/ecp_nistz256.c", line 1252.70: 1506-1298 (W) The subscript 7 is out of range. The valid range is 0 to 3.
[...]

PR #112, opened 2018-08-28
OpenSSL crypto support for OpenJ9-OpenJDK build

mbvreddy:
Issue #115

Add support to use OpenSSL native implementation for some Cryptography Algorithms in OpenJ9 build. By default, this support is disabled (as of now). Below system properties are in place to control it

-Djdk.nativeCrypto=true
-Djdk.nativeDigest=true
-Djdk.nativeCBC=true
-Djdk.nativeGCM=true

New JNI library **jnicrypto** is created as part of this and is bundled with final JDK. However, there is no plan to ship OpenSSL libraries with OpenJ9. 

This is enabled on all platforms except for zOS and Windows.

PR #124, opened 2018-10-08
Build with OpenSSL v1.1.1 on AIX

mbvreddy:
This PR is for issue #123 

#112  enabled OpenSSL support in openJ9-openjdk builds which also allow bundling of OpenSSL crypto library with JDK build. However build fails with `--enable-openssl-bundling` option as OpenSSL 1.1.1 changed the library name to align with the natural way of handling shared library on AIX. Corresponding openssl PR for reference - https://github.com/openssl/openssl/pull/6487

As a result, we now have to link and bundle libcrypto.a instead of libcrypto.so.1.1.

In this fix, 
- we copy libcrypto.a to jre/lib/ppc64 directory if built with OpenSSL v1.1.1 and libcrypto.so to jre/lib/ppc64 if built with OpenSSL 1.1.0x
- remove linking against `ssl` on all platforms as our implementation use only `crypto` library

### DIFF
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4336,7 +4336,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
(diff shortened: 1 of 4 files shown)


andrew-m-leonard:
jenkins compile aix

PR #125, opened 2018-10-09
Avoid using `--exclude-vcs` which isn't universally supported

keithc-ca:
This eliminates the need to install a special version of tar on macosx.


pshipton:
jenkins compile win,plinux,aix

PR #132, opened 2018-10-11
No warning message when failed to load native crypto library

### LINKED ISSUE #127: No warning message when failed to load native crypto library and falls back to java crypto implementation.
If the native crypto library is not available in the library path, then  System.loadLibrary fails with an UnsatisfiedLinkError. In this situation, there is a fall back mechanism to the java crypto implementation by catching the exception. However, there is no warning messages issued while falling back and hence the user do not know the native crypto implementation is not used.

enasser:
…s back to java crypto implementation

The following changes are made as part of this changeset:
   - Added a new warning message when native crypto library load failed and fallback to java crypto implementation.
   - The native crypto property is checked before loading the native crypto library. This will avoid unnecessary
     loading of the native library when user did not specify the native crypto property. Also, it avoids throwing
     the warning message when user did not specify the native crypto property.
   - Refine the use of native crypto property flags.
   - Use a method to check whether native crypto library is loaded successfully instead of accessing the class
     variable directly.

fixes #127

### DIFF
--- a/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -25,21 +25,25 @@ import java.security.*;
 
 public class NativeCrypto {
 
-    public static boolean isLoaded = false;
+    private static boolean loaded = false;
 
     static {
         AccessController.doPrivileged(new PrivilegedAction<Void>() {
             public Void run() {
                 try {
                     System.loadLibrary("jncrypto"); // check for native library
-                    isLoaded = true;
+                    loaded = true;
                 } catch (UnsatisfiedLinkError usle) {
-                    isLoaded = false;
+                    loaded = false;
                 }
                 return null;
             }
         });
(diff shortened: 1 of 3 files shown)


pshipton:
jenkins test sanity plinux,aix

pshipton:
I note this code does not allow for options like `-Djdk.nativeCrypto=true -Djdk.nativeCBC=false`

pshipton:
jenkins test sanity win,xlinux

pshipton:
jenkins test sanity win32

andrew-m-leonard:
I'll let Peter merge once builds clean and he's ok with changes...

enasser:
> I note this code does not allow for options like -Djdk.nativeCrypto=true -Djdk.nativeCBC=false

Yes @pshipton. Current code does not allow enable all and disable few. Currently only three (Digest, CBC & GCM) are supported and hence did not see the need of enabling all and disable few. However, we can consider including that feature when more algorithms are supported.

pshipton:
The aix failure is caused by https://github.com/eclipse/openj9/issues/3248

PR #136, opened 2018-10-12
WIP: AccessControlException from CipherCore when Security manager enabled

### LINKED ISSUE #135: AccessControlException from CipherCore when Security manager enabled
The following test cases are failed with openssl:
com/sun/crypto/provider/Cipher/AES/TestAESCiphers/TestAESWithProviderChange.java.TestAESWithProviderChange
com/sun/crypto/provider/Cipher/AES/TestAESCiphers/TestAESWithRemoveAddProvider.java.TestAESWithRemoveAddProvider

https://hyc-runtimes-jenkins.swg-devops.com/view/Test_security/job/openjdk8_j9_jdksecurity1test_x86-64_linux/18/

Caused by: java.security.AccessControlException: Access denied ("java.lang.RuntimePermission" "accessClassInPackage.sun.security.action")
	at java.security.AccessController.throwACE(AccessController.java:176)
	at java.security.AccessController.checkPermissionHelper(AccessController.java:237)
	at java.security.AccessController.checkPermission(AccessController.java:373)
	at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
	at java.lang.SecurityManager.checkPackageAccess(SecurityManager.java:1564)
	at java.lang.J9VMInternals$2.run(J9VMInternals.java:268)
	at java.security.AccessController.doPrivileged(AccessController.java:673)
	at java.lang.J9VMInternals.checkPackageAccess(J9VMInternals.java:263)
	at com.sun.crypto.provider.CipherCore.<clinit>(CipherCore.java:1231)

The issue is that from com.sun.crypto.provider.CipherCore, we are trying to access the class sun.security.action.GetPropertyAction. CipherCore is loaded with extension classloader and GetPropertyAction is loaded with boot class loader. This will result in a security exception as package.access restriction is defined in java.security. We can workaround by providing the RuntimePermission in the policy file. However, that is not a good approach as it is against the definition in java.security.. We need to read the property differently to avoid the issue..

Also, it makes more sense to have the value of jdk.nativeCrypto as 'yes' or 'no' instead of 'true' or 'false' as we need to change the default to enable native crypto when code stabilizes and then jdk.nativeCrypto=no instead of jdk.nativeCrypto=false to disable native crypto makes more sense.

enasser:
The following changes are included in this changeset:
   - Removed the usage of sun.security.action.GetPropertyAction in
     com.sun.crypto.provider.CipherCore as it caused
     java.security.AccessControlException.
   - Change the value of native crypto properties from TRUE|FALSE to
     YES|NO as it makes more sense for end user to specify
     jdk.nativeCrypto=NO insted of jdk.nativeCrypto=FALSE.
   - Made the property value case insensitive so that user can
     specify 'YES' or 'Yes' or 'yes'.
   - Allow user to specify enable all native crypto and disable
     specific crypto

fixes #135

### DIFF
--- a/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -39,7 +39,7 @@ import java.security.spec.*;
 import javax.crypto.*;
 import javax.crypto.spec.*;
 import javax.crypto.BadPaddingException;
-import sun.security.action.GetPropertyAction;
+
 import jdk.crypto.jniprovider.NativeCrypto;
 
 /**
@@ -73,8 +73,6 @@ final class CipherCore {
      * 'jdk.nativeCrypto' is used to enable all native cryptos (Digest,
      * CBC and GCM).
      */
-    private static boolean useNativeCrypto = false;
-
     private static boolean useNativeCBC = false;
 
     private static boolean useNativeGCM = false;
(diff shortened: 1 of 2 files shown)


pshipton:
@enasser an AccessController.doPrivileged() is still required around System.getProperty() in order for this api to not throw an AccessControlException (SecurityException) when a SecurityManager is installed.

enasser:
Thank you @pshipton for your inputs. I was initially thinking it may already have the privilege while invoking those methods as I could see other System.getProperty() without AccessController.doPrivileged() in those classes and application cannot directly access those methods. Hence, I thought of adding it if required after testing to avoid the over head of additional doPrivileged. However, on second thought I feel it's better to add at least for CipherCore as it's an API and may directly access from application. Whether JIT has the intelligence to skip doPrivileged if the thread already has the privilege.

pshipton:
obsoleted by #145

PR #134, opened 2018-10-12
Correct --with-openssl lib location

andrew-m-leonard:
Signed-off-by: andrew-m-leonard <andrew_m_leonard@uk.ibm.com>

PR #137, opened 2018-10-12
Move CONFIGURE-OPENSSL to CUSTOM-LATE-HOOK to resolve library prefix &

andrew-m-leonard:
Moved CONFIGURE-OPENSSL to CUSTOM-LATE-HOOK so as to resolve SHARED-LIBRARY-SUFFIX...


andrew-m-leonard:
jenkins compile xLinux

pshipton:
What is the problem this is resolving? i.e. what isn't working that is fixed by this change?

mbvreddy:
This resolves LIBRARY_PREFIX and SHARED_LIBRARY_SUFFIX successfully in #134 so that openssl check with --with-openssl=/path/to/custom/openssl would be successful.

change looks good to me.

andrew-m-leonard:
The provided openssl lib still could not be resolved as the script is looking for:
```
if test -s "$OPENSSL_DIR/lib/${LIBRARY_PREFIX}crypto${SHARED_LIBRARY_SUFFIX}.1.1"; then
```
however SHARED_LIBRARY_SUFFIX is defined later on....!

PR #140, opened 2018-10-12
Fix crypto lib path when --with-openssl=/path/to/pre-built/openssl is used

### LINKED ISSUE #139: Build fails with configure option --with-openssl=/path/to/pre-built/openssl --enable-openssl-bundling  on unix platforms
when --with-openssl=/path/to/pre-built/openssl --enable-openssl-bundling configure options are provided, build will copy libcrypto.so.1.1 to JDK directory. However, when directory structure of the openssl directory is different, configure couldn't check for openssl and thereby it fails. The same is fixed with #134. With this configure step is successful and fails when copying OpenSSL crypto library to JDK folder.

It is also required to fix the path from where OpenSSL crypto library is copied in the same way as #134.

mbvreddy:
fixes #139 

This PR fix the path of OpenSSL crypto library from which build copies it to JDK. This is very similar to PR #134 

This fix checks for libcrypto.so.1.1 in both locations $OPENSSL_DIR/ and $OPENSSL_DIR/lib and set OPENSSL_BUNDLE_LIB_PATH accordingly. 

### DIFF
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4336,7 +4336,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
-DATE_WHEN_GENERATED=1539348366
+DATE_WHEN_GENERATED=1539368198
 
 ###############################################################################
 #
--- a/jdk/make/closed/autoconf/custom-hook.m4
+++ b/jdk/make/closed/autoconf/custom-hook.m4
@@ -479,6 +479,9 @@ AC_DEFUN([CONFIGURE_OPENSSL],
         else
           BUILD_OPENSSL=yes
         fi
+        if test "x$BUNDLE_OPENSSL" = xyes; then
+          OPENSSL_BUNDLE_LIB_PATH=$OPENSSL_DIR
+        fi
         AC_MSG_RESULT([yes])
       else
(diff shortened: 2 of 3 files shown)


mbvreddy:
I've built with this change on zLinux locally with configure option --with-openssl=/root/openssl-1.1.1 and job is successfull. I could verify that libcrypto.so.1.1 gets copied to JDK folder.

```
root@ojdk-zlinux1:~/openj9/mbvreddy/openj9-openjdk-jdk8/build/linux-s390x-normal-server-release/images/j2sdk-image/jre/lib/s390x# ldd libjncrypto.so 
	libcrypto.so.1.1 => /root/openj9/mbvreddy/openj9-openjdk-jdk8/build/linux-s390x-normal-server-release/images/j2sdk-image/jre/lib/s390x/./libcrypto.so.1.1 (0x000003ff87580000)
	libc.so.6 => /lib/s390x-linux-gnu/libc.so.6 (0x000003ff87300000)
	libdl.so.2 => /lib/s390x-linux-gnu/libdl.so.2 (0x000003ff87280000)
	libpthread.so.0 => /lib/s390x-linux-gnu/libpthread.so.0 (0x000003ff87200000)
	/lib/ld64.so.1 (0x000003ff87900000)
```

pshipton:
jenkins compile zlinux depends eclipse/openj9#3250

pshipton:
jenkins compile win

pshipton:
Although the zlinux compile seems unfinished, the compile itself completed and ran -version.

PR #141, opened 2018-10-15
OpenJ9 JDK8 OSX Support

babsingh:
**1) sed on Mac OS doesn't recognize \r as a special character**

  \r - carriage return.

  tr on Mac OS recognizes \r. Thus, replaced sed with tr.

**2) Add jvm.cfg for Mac OSX**


babsingh:
fyi - @pshipton @DanHeidinga @groeges

pshipton:
jenkins compile win,xlinux,aix

PR #144, opened 2018-10-17
JDK8 OSX Support

babsingh:
**1) Use custom compiler to build OpenJ9 JDK8 on OSX**

On OSX, JDK8 is built using Xcode4.6.3 + gcc4.2. These tools won't
build OpenJ9. Tool requirements for OpenJ9 are: Xcode7.2.1 + gcc4.9.
But, Xcode7.2.1 + gcc4.9 can't be used to build JDK components.

This feature will allow a user to specify custom tools to build
OpenJ9. Usage on OSX:
```
export OPENJ9_USE_CUSTOM_COMPILER=1
export OPENJ9_DEVELOPER_DIR="/Applications/Xcode7/Xcode.app/Contents/Developer"
export OPENJ9_CC="/usr/local/bin/gcc-4.9"
export OPENJ9_CXX="/usr/local/bin/g++-4.9"
```
**2) Set MACOSX_DEPLOYMENT_TARGET and map memory below 4 GB**

- Setting `MACOSX_DEPLOYMENT_TARGET` environment variable will let OpenJ9
executables to run on OSX versions older than the OSX version on which
OpenJ9 was built. `MACOSX_DEPLOYMENT_TARGET` acts similar to 
`-mmacosx-version-min=version` compiler option. If both the compiler option 
is specified and the environment variable is set, then the compiler option will take
precedence. Here, `MACOSX_DEPLOYMENT_TARGET` environment variable and
the compiler option will point to the same version. The environment
variable is defined to support dependencies where the compiler option
is not applied. Issue: eclipse/openj9#3244

- Append "-pagezero_size 0x1000" to LDFLAGS_JDKEXE on OSX 64-bit
compressedrefs build. This allows memory to be mapped below 4GB.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -373,6 +373,16 @@ define create_or_update
 	fi
 endef
 
+ifeq (1,$(OPENJ9_USE_CUSTOM_COMPILER))
+  SETUP_CUSTOM_COMPILER_ENV_VARS := \
+	DEVELOPER_DIR="$(OPENJ9_DEVELOPER_DIR)" \
+	CC="$(OPENJ9_CC)" \
+	CXX="$(OPENJ9_CXX)" \
+	#
+else
+  SETUP_CUSTOM_COMPILER_ENV_VARS :=
+endif
+
 run-preprocessors-j9 : stage-j9
 	@$(ECHO) Ensuring version information is up-to-date
 	$(call create_or_update, \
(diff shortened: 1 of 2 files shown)


babsingh:
fyi - @DanHeidinga 

related to https://github.com/eclipse/openj9/issues/36

closed/OpenJ9.gmk
```diff
@@ -373,6 +373,25 @@ define create_or_update
...
+ifeq (1,$(OPENJ9_USE_CUSTOM_COMPILER))
+  SETUP_CUSTOM_COMPILER_ENV_VARS := \
+	DEVELOPER_DIR_OLD="$(DEVELOPER_DIR)" \
+	CC_OLD="$(CC)" \
+	CXX_OLD="$(CXX)" \
+	DEVELOPER_DIR="$(OPENJ9_DEVELOPER_DIR)" \
+	CC="$(OPENJ9_CC)" \
+	CXX="$(OPENJ9_CXX)" \
+	#
+  CLEANUP_CUSTOM_COMPILER_ENV_VARS := \
```
keithc-ca: There's no need to capture the old values of these environment variables: they're only set in a sub-shell created by make.

babsingh: @keithc-ca i have removed redundant `_OLD` variables.

keithc-ca:
`OPENJ9_USE_CUSTOM_COMPILER` should not be set in any current build: merging without a PR build.

PR #145, opened 2018-10-18
Set OpenSSL to be used by default

### LINKED ISSUE #135: AccessControlException from CipherCore when Security manager enabled

ashbm5:
This PR includes two things:
1. Set OpenSSL native crypto library to be used by default.
2. Fix AccessControlException from CipherCore when Security manager enabled

also added is the property jdk.nativeCryptoTrace that will turn on printing traces messages relevant to the native crypto function.

Fixes #135

### DIFF
--- a/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -39,8 +39,9 @@ import java.security.spec.*;
 import javax.crypto.*;
 import javax.crypto.spec.*;
 import javax.crypto.BadPaddingException;
-import sun.security.action.GetPropertyAction;
 import jdk.crypto.jniprovider.NativeCrypto;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 
 /**
  * This class represents the symmetric algorithms in its various modes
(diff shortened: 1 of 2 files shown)


pshipton:
jenkins compile zlinux

pshipton:
jenkins compile zlinux

pshipton:
jenkins compile plinux

PR #152, opened 2018-10-23
Remove tools.jar from bootclasspath

keithc-ca:
It doesn't exist within the jre folder named by the expanded path.


pshipton:
jenkins test all xlinux

PR #154, opened 2018-10-23
Adjust makefiles to avoid duplicate classes in bootclasspath jars

### LINKED ISSUE #142: Charset classes duplicated in rt.jar charsets.jar
sun/awt/motif/X11GB2312$Decoder.class sun/awt/motif/X11GB2312$Encoder.class sun/awt/motif/X11GBK$Encoder.class sun/awt/motif/X11KSC5601$Decoder.class sun/awt/motif/X11KSC5601$Encoder.class are duplicated between rt.jar, charsets.jar, and localedata.jar

The problem is due to the CreateJars.gmk \$$$$ escaping and the extensions implementation of ListPathsSafely which results in $$ in the class name output in the build exclude file, rather than just $.
The original implementation works for this.

keithc-ca:
* the changes made to make/common/MakeBase.gmk to enable DDR had unintended consequences; they have been reverted
* the makefile to compile generated DDR classes has been reworked so it is not dependent on `SetupJavaCompilation` which doesn't handle `$` properly

Fixes #142
Fixes eclipse/openj9#3393

### DIFF
--- a/closed/make/DDR-jar.gmk
+++ b/closed/make/DDR-jar.gmk
@@ -19,7 +19,7 @@
 all :
 
 ifeq (,$(wildcard $(SPEC)))
-  $(error BuildDDR.gmk needs SPEC set to a proper spec.gmk)
+  $(error DDR-jar.gmk needs SPEC set to a proper spec.gmk)
 endif
 
 include $(SPEC)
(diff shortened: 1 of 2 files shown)


pshipton:
This is a change to the OpenJDK code, yet OpenJDK doesn't have the same problem. It seems the fix should be elsewhere so there isn't a patch to OpenJDK code required.

keithc-ca:
OpenJDK apparently worked around the issue that motivated the earlier change to `ListPathsSafely`: I'll explore reverting that earlier change and adapting the DDR makefiles accordingly.

keithc-ca:
The copyright failure for `make/common/MakeBase.gmk` is expected: it is reverted to the version from openjdk.

pshipton:
jenkins test all xlinux,win

pshipton:
jenkins test sanity win32

pshipton:
jenkins compile win32

pshipton:
Stopped the compile job as it is unneeded.

PR #160, opened 2018-10-25
Exclude OpenJ9 management extensions from compact profiles 1 & 2

keithc-ca:
Fixes https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/155.

### DIFF
--- a/closed/make/CreateJars.gmk
+++ b/closed/make/CreateJars.gmk
@@ -30,6 +30,22 @@ $(eval $(call SetupArchive,BUILD_DTFJVIEW_JAR, , \
 	JAR := $(IMAGES_OUTPUTDIR)/lib/ext/dtfjview.jar, \
 	SKIP_METAINF := true))
 
+TRACEFORMAT_TYPES := \
+	com/ibm/jvm/Debug.class \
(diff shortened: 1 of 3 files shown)


closed/make/CreateJars.gmk
```diff
@@ -86,22 +86,99 @@ RT_JAR_EXCLUDES += \
...
+	com/ibm/jvm/dtfjview \
+	com/ibm/jvm/format \
+	com/ibm/jvm/j9 \
+	com/ibm/jvm/trace \
+	com/ibm/jvm/traceformat \
 	com/ibm/tools/attach/attacher \
 	#
+
+OPENJ9_MANAGEMENT_TYPES := \
+	com/ibm/java/lang/management/internal/BufferPoolMXBeanImpl.class \
```
pshipton: Can we use wildcards to include entire packages? A hard code list of classes will likely result in problems when new classes are added.

keithc-ca: Yes, we ought to be do that.

keithc-ca:
Updated as suggested.

pshipton:
jenkins test sanity zlinux

PR #172, opened 2018-10-30
Include jextract in JRE image

keithc-ca:
Having `jextract` available in a JRE makes it easier for users to collect information that will help debug problems.


pshipton:
jenkins signed off by check

pshipton:
I'm thinking this change should also go into the openj9-0.11 branch.

keithc-ca:
> I'm thinking this change should also go into the openj9-0.11 branch.

Agreed: I'll put together another PR for that.

PR #194, opened 2018-11-02
Use ppc64le for CPU name on little-endian POWER systems

### LINKED ISSUE #190: vm directory for PPC LE is ppc64 but should be ppc64le
In OpenJ9 PPC LE builds, the vm directory path uses ppc64, while it should be ppc64le. The OpenJDK builds use ppc64le.

keithc-ca:
Fixes #190.


keithc-ca:
Fixed copyright notice.

common/autoconf/platform.m4
```diff
@@ -67,7 +70,7 @@ AC_DEFUN([PLATFORM_EXTRACT_VARS_FROM_CPU],
       VAR_CPU_ENDIAN=big
       ;;
     powerpc64le)
-      VAR_CPU=ppc64
+      VAR_CPU=ppc64le
```
pshipton: Any idea why OpenJDK builds use a ppc64le directory already without this fix?

keithc-ca: No idea.

pshipton: I have concerns about merging this. As OpenJDK uses a ppc64le directory without this change, it implies there is something wrong with the OpenJ9 build process that can be fixed without patching the OpenJDK code.

keithc-ca: The new version I pushed earlier today changes only 'closed' (and `generated-configure.sh`) files.

pshipton:
It seems these fixes should be split into multiple PRs. All the fix details are in the commit comments. This PR only has a one line description.

keithc-ca:
The changes are inter-related: I've updated the description with more details.

pshipton:
jenkins compile win,xlinux,plinux,aix

pshipton:
Keith is looking at the failure

keithc-ca:
This is essentially my first proposal; it makes changes similar to patches applied to openjdk by AdoptOpenJDK (see https://github.com/AdoptOpenJDK/openjdk-build/pull/707).

pshipton:
jenkins test all plinux

pshipton:
Seems there are more changes required to the builds, the build script cannot find the JVM to run -version.

keithc-ca:
Jenkins test all plinux jdk8 depends eclipse/openj9#3649

keithc-ca:
See the successful JDK8 builds linked to https://github.com/eclipse/openj9/pull/3649.

PR #202, opened 2018-11-08
Remove OpenJDK/Hotspot attach api classes

ChengJin01:
The change is to remove these api classes
and stop them from generating attach.dll
on Windows as these classes are replaced
with the implementation in OpenJ9.

Related to https://github.com/eclipse/openj9/issues/3441

Signed-off-by: CHENGJin <jincheng@ca.ibm.com>

### DIFF
--- a/closed/make/CompileJavaClasses.gmk
+++ b/closed/make/CompileJavaClasses.gmk
@@ -1,0 +1,27 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+# 
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  
+#
+# IBM designates this particular file as subject to the "Classpath" exception
+# as provided by IBM in the LICENSE file that accompanied this code.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+# 
+# ===========================================================================
+
+# Replace attach API classes on HotSpot with OpenJ9 implementation.
+ifeq ($(OPENJDK_TARGET_OS), windows)
+  EXFILES+=sun/tools/attach/WindowsAttachProvider.java \
+           sun/tools/attach/WindowsVirtualMachine.java
+endif
--- a/closed/make/CompileNativeLibraries.gmk
+++ b/closed/make/CompileNativeLibraries.gmk
@@ -1,0 +1,26 @@
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+# 
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  
+#
+# IBM designates this particular file as subject to the "Classpath" exception
+# as provided by IBM in the LICENSE file that accompanied this code.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+# 
+# ===========================================================================
+
+# Avoid generating attach.dll from the natives of attach APIs on Hotspot.
+ifeq ($(OPENJDK_TARGET_OS), windows)
+  BUILD_LIBRARIES := $(filter-out $(BUILD_LIBATTACH), $(BUILD_LIBRARIES))
+endif


ChengJin01:
Reviewer: @pshipton

pshipton:
@ChengJin01 what is the issue where we discussed making this change? Please refer to it in the PR description.

keithc-ca:
You should not need to modify or delete any openjdk files to achieve the goal of this PR.

Add a custom `CompileNativeLibraries.gmk` file which says:
```
ifeq ($(OPENJDK_TARGET_OS), windows)
  BUILD_LIBRARIES := $(filter-out $(BUILD_LIBATTACH), $(BUILD_LIBRARIES))
endif
```
to remove the dependency on building the attach library.

The deletions can be accomplished by adding empty (except for the copyright notice) files with the right names in `closed/overrides/jdk/src/windows/...` (with the above, you may not need overrides for the `*.c` files).

ChengJin01:
> what is the issue where we discussed making this change? Please refer to it in the PR description.

The PR was done based on https://github.com/eclipse/openj9/issues/3441#issuecomment-435362195 so as to narrow down the scope of problem.

ChengJin01:
> You should not need to modify or delete any openjdk files to achieve the goal of this PR. ...
Agreed, will update against the suggestion above.

ChengJin01:
I created a customized CompileNativeLibraries.gmk at /openj9-openjdk-jdk8/closed/make/lib and empty .java/.c files at closed/overrides/jdk/src/windows while still keeping the original source files.

```
closed/make/lib
$ ls -l
total 8
-rw-rw-r--+ 1 Administrator  None 1184 Nov  5 10:05 CoreLibraries.gmk
-rwxrwxr-x+ 1 Administrators None 1222 Nov 12 13:25 ServiceabilityLibraries.gmk

/closed/overrides/jdk/src/windows/classes/sun/tools/attach
$ ls -l
total 8
-rwxrwxr-x+ 1 Administrators None 1151 Nov 12 13:30 WindowsAttachProvider.java
-rwxrwxr-x+ 1 Administrators None 1151 Nov 12 13:30 WindowsVirtualMachine.java

closed/overrides/jdk/src/windows/native/sun/tools/attach
$ ls -l
total 8
-rwxrwxr-x+ 1 Administrators None 1151 Nov 12 13:32 WindowsAttachProvider.c
-rwxrwxr-x+ 1 Administrators None 1151 Nov 12 13:32 WindowsVirtualMachine.c

```

However, attach.dll still occurred and the original class files still show up in tools.jar after compilation (I already checked)
```
windows-x86_64-normal-server-release/images/j2sdk-image
$ find . -name attach.*
./jre/bin/attach.diz
./jre/bin/attach.dll

windows-x86_64-normal-server-release/images/j2sdk-image/lib/tools/sun/tools/attach
$ .../bin/javap -v WindowsVirtualMachine.class
Classfile .../windows-x86_64-normal-server-release/images/j2sdk-image/lib/tools/sun/tools/attach/WindowsVirtualMachine.class
  Last modified Nov 12, 2018; size 3217 bytes
  MD5 checksum 65361f82348fb4881c64d5b7c80dcbd3
  Compiled from "WindowsVirtualMachine.java"
public class sun.tools.attach.WindowsVirtualMachine extends sun.tools.attach.HotSpotVirtualMachine
  minor version: 0
  major version: 52
```

There might be something incorrect with the proposed changes.

keithc-ca:
Please remove 'WIP' from the title once you have this working.

ChengJin01:
After tried a couple of options, I successfully removed the native/java files without modifying the openjdk files (changes already updated and verified with the compilation results)

1) closed/make/CompileNativeLibraries.gmk to remove the native code
It works good as expected (previously put in the wrong place)

2) closed/make/CompileJavaClasses.gmk to remove the java code
There is no need to override any file as **EXFILES** is defined in CompileJavaClasses.gmk to exclude requested files during compilation.

The compilation results are as follows:
```
/openj9-openjdk-jdk8/build
$ find . -name attach.dll
(nothing shows up which is correct /already checked images\j2sdk-image\jre\bin & jdk\bin)

/openj9-openjdk-jdk8/build/windows-x86_64-normal-server-release/images/lib/tools/sun/tools/attach (extracted from tools.jar)
$ ls -l
total 20
-rwxrwxr-x+ 1 Administrators None  548 Nov 15 22:22 'HotSpotAttachProvider$HotSpotVirtualMachineDescriptor.class'
-rwxrwxr-x+ 1 Administrators None 3538 Nov 15 22:22  HotSpotAttachProvider.class
-rwxrwxr-x+ 1 Administrators None 8440 Nov 15 22:22  HotSpotVirtualMachine.class
(WindowsAttachProvider & WindowsVirtualMachine didn't get compiled)

```

pshipton:
@keithc-ca can you please take a look. 

I want to check (later) if there are other attach classes that should also be removed, but they can just be added to the list.

keithc-ca:
I was wondering if the attach library should just be omitted on all platforms.
I noticed that on Linux it's in the JDK, but only the `.diz` file is in the JRE!
```sh
$ find images/j2* -name '*attach*' -type f
images/j2re-image/lib/amd64/libattach.diz
images/j2sdk-image/jre/lib/amd64/libattach.diz
images/j2sdk-image/jre/lib/amd64/libattach.so
```
I would expect `libattach.so` should be included in the JRE as well.

pshipton:
I'm ok with omitting the attach library on all platforms, it isn't used on any platform.

keithc-ca:
> I'm ok with omitting the attach library on all platforms, it isn't used on any platform.

I think we should defer that to a separate PR.

pshipton:
lgtm

ChengJin01:
Will investigate to see what happens there.

ChengJin01:
Looking at the test code at /Java8andUp/src/org/openj9/test/java/lang/Test_ClassLoader.java
```
	public void test_callNativesOnNewClassLoaders() {
		// Ensure calling these natives on a new ClassLoader() does not crash
                ...
		try {
			/*
			 * It should be fine to re-use a previous loader, as the previous
			 * tests should not have created any native structures, but create a
			 * new loader anyway in case something changes in the future.
			 */
			LoaderSubclass loader2 = new LoaderSubclass();
			/*
			 * This native will allocate native structures. This test is
			 * somewhat fragile. Load an existing library which exists but is
			 * not already loaded by another classloader. If it starts failing,
			 * find another library to load.
			 */
			loader2.loadLibrary("attach");  <------------------
		} catch (UnsatisfiedLinkError e) {
			e.printStackTrace();
			Assert.fail("expected to find library");
		}
```
It seems the failing test was trying to load attach.dll for general purpose. Given that there is no specific requirement on the natives being loaded,  I will randomly choose another one to replace attach.dll (e.g. awt.dll) to see whether it works.

keithc-ca:
`awt` may not be the best choice: I suspect it might be absent in 'headless' configurations.

ChengJin01:
> The test code says it needs a library that isn't already loaded. Perhaps the test needs one of its own?

attach.dll is definitely not one of test libraries and I didn't see there is any restriction on which library to be loaded in this test.  In addition, I already tried with awt.dll in personal build and the result is OK on Windows.

keithc-ca:
The library loaded by that test must be available on all platforms and in all configurations that we would want to test. As I said, that might include a 'headless' configuration that may not include a library for `awt`. Testing only on Windows is not sufficient. I'm not sure what the best alternate library might be, but `attach` definitely won't work, `awt` might not work. Please share a link here to a PR that makes appropriate changes to the test.

ChengJin01:
There are two options for us:
1) find out a libraray which isn't already loaded
2) create a library (added to the JDK) only used for this test (Pete objected to this option)

The PR for the test case changes has been created at https://github.com/eclipse/openj9/pull/3720

Issue #201, opened 2018-11-08
Don't need jextract in bin


pshipton:
#180 added jextract to the jre/bin directory, but I notice it's also in the bin directory. Since it's not a development tool, it doesn't need to be in the bin directory.

pshipton:
@keithc-ca

keithc-ca:
There's already a large number of executables in both directories (e.g. java, javaw, javaws, pack200, unpack200, etc.): is this a significant concern?

pshipton:
Significant no. Just wrong.

PR #205, opened 2018-11-09
Only include jextract in jre/bin

keithc-ca:
Fixes https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/201.


pshipton:
jenkins test sanity zlinux,win

PR #206, opened 2018-11-09
Change MACOSX_VERSION_MIN to 10.9.0 for OpenJ9 JDK8

babsingh:
For OpenJ9 JDK8 on OSX, the JIT can't be built with MACOSX_VERSION_MIN
<= 10.8.0 using Xcode7 + clang. JIT needs libc++ which is only available when
MACOSX_VERSION_MIN is set to 10.9.0 or greater. With MACOSX_VERSION_MIN
<= 10.8.0, the Xcode7 build tools only provide an old version of stdlibc++, which
can't be used to build the JIT.

Thus, changing MACOSX_VERSION_MIN from 10.7.0 to 10.9.0 in order to
support the JIT when building OpenJ9 JDK8 on OSX using Xcode7 + clang.

Invoked autogen.sh.

Signed-off-by: Babneet Singh sbabneet@ca.ibm.com

### DIFF
--- a/common/autoconf/flags.m4
+++ b/common/autoconf/flags.m4
@@ -23,6 +23,12 @@
 # questions.
 #
 
+#
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+#
+
 AC_DEFUN_ONCE([FLAGS_SETUP_INIT_FLAGS],
 [
   # Option used to tell the compiler whether to create 32- or 64-bit executables
@@ -574,7 +580,7 @@ AC_DEFUN_ONCE([FLAGS_SETUP_COMPILER_FLAGS_FOR_JDK],
       # newer than the given OS version and makes the linked binaries compatible 
       # even if built on a newer version of the OS.
       # The expected format is X.Y.Z
-      MACOSX_VERSION_MIN=10.7.0
+      MACOSX_VERSION_MIN=10.9.0
       AC_SUBST(MACOSX_VERSION_MIN)
     
       # The macro takes the version with no dots, ex: 1070
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -3685,6 +3685,12 @@ ac_configure="$SHELL $ac_aux_dir/configure"  # Please don't use this var.
 # questions.
 #
 
+#
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+#
+
 
 
 
@@ -4336,7 +4342,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
-DATE_WHEN_GENERATED=1541112610
+DATE_WHEN_GENERATED=1542039989
 
 ###############################################################################
 #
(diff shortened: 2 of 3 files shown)


babsingh:
related to https://github.com/eclipse/openj9/issues/3496 and https://github.com/eclipse/openj9/issues/36.

babsingh:
fyi - @DanHeidinga @pshipton @nbhuiyan

DanHeidinga:
https://en.wikipedia.org/wiki/MacOS_version_history#Version_10.7:_"Lion"

`10.7.0` was released July 20, 2011 
`10.9.0` was released October 22, 2013

DanHeidinga:
@babsingh Can you fix the copyrights so the check will pass?

keithc-ca:
Building Java 11 doesn't seem to have this problem. In that version I only see a dependency on `libc++.1.dylib` and no dependency on any version `libstdc++` as discussed in https://github.com/eclipse/openj9/issues/3496. I would hope we can accomplish the same for Java 8.

babsingh:
> Can you fix the copyrights so the check will pass?

Currently, [common/autoconf/flags.m4](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/02e895a185de5ab5964b3ea0a6e4b87c890684b8/common/autoconf/flags.m4#L2) has an Oracle copyright with the correct year (2018). Our copyright check does: `grep -qE \(c\) Copyright IBM Corp. ([0-9]{4}), <FILE>`. In other files such as [NativeGaloisCounterMode.java](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/closed/adds/jdk/src/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java), IBM's copyright is added below the Oracle copyright. I have done the same for `flags.m4`.

Running `autogen.sh` creates two `generated-configure.sh` (common/closed) scripts. The copyright check also fails for the two `generated-configure.sh` scripts. In the past, copyright check failures due to running `autogen.sh` were ignored. I am guessing that it still holds true.

> Building Java 11 doesn't seem to have this problem. In that version I only see a dependency on libc++.1.dylib and no dependency on any version libstdc++ as discussed in eclipse/openj9#3496. I would hope we can accomplish the same for Java 8.

With this change, OpenJ9 JDK8 will have no dependency on any `libstdc++` version. JIT will only have a dependency on `libc++.1.dylib` similar to OpenJ9 JDK11. 

**Side notes:**
- OpenJ9 JDK11 has `MACOSX_VERSION_MIN` set to `10.9.0`.
- Also, I think the behaviour regarding availability of `libc++` varies depending upon OSX build tools. Xcode7 provides `libc++` support for `MACOSX_VERSION_MIN` >= `10.9.0`. But, Xcode9 provides `libc++` support for `MACOSX_VERSION_MIN` >= `10.7.0`. This observation was made while building OpenJ9 JDK9 and OpenJ9 JDK10; both use `MACOSX_VERSION_MIN` = `10.7.0`. No dependency on `libstdc++` was seen when building OpenJ9 JDK9 and OpenJ9 JDK10 with Xcode9 and `MACOSX_VERSION_MIN` = `10.7.0`.

DanHeidinga:
Jenkins test sanity osx jdk11

DanHeidinga:
We don't have a JDK8 PR build yet so this will have to be merged without testing beyond a compile on another platform

DanHeidinga:
Jenkins compile xlinux jdk8

DanHeidinga:
@keithc-ca Any further comments?  Otherwise I'll merge this

keithc-ca:
No need to re-run PR testing for the most recent change: merging.

PR #207, opened 2018-11-12
Use correct version of JCL library

keithc-ca:
Java 8 should be using `libjclse7b_29.so`, not `libjclse9_29.so` (which is for Java 9).

Required by https://github.com/eclipse/openj9/pull/3638.


keithc-ca:
Jenkins sanity xlinux jdk8

keithc-ca:
Jenkins test sanity xlinux jdk8

Issue #189, opened 2018-11-02
OpenJ9 builds do not put libjsig.so / jsig.dll in the correct places


pshipton:
Windows:
bin\jsig.dll

Linux/AIX:
lib/libjsig.so

The jsig file does not need to be in the compressedrefs/default directory.

keithc-ca:
Given that https://github.com/eclipse/openj9/issues/2100 has been closed, I expect the answer to this will be to back-port the makefile patterns used in Java 11+ which will mean none of the test artifacts will be included in a JRE or JDK image.

keithc-ca:
Fixed in #211; verified via https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u-2019-03-14-16-57/OpenJDK8U-jdk_x64_windows_openj9_2019-03-14-16-57.zip

Issue #188, opened 2018-11-02
OpenJ9 builds do not include jsig.lib on Windows


pshipton:
The jsig.lib file should be included in the lib directory.

keithc-ca:
Verified using https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u202-b08_openj9-0.12.1/OpenJDK8U-jdk_x64_windows_openj9_8u202b08_openj9-0.12.1.zip

PR #211, opened 2018-11-16
Put jsig libraries in the correct places

keithc-ca:
* On Windows the shared library should be in `bin`, `bin/j9vm` and `bin/server` (in an SDK, the static library should be in `lib`).
* On other platforms, the shared library should be in `lib/{arch}`, `lib/{arch}/j9vm` and `lib/{arch}/server`.


keithc-ca:
This should address #188 and #189.

pshipton:
jenkins test sanity win,xlinux

PR #213, opened 2018-11-19
[WIP] SSL Connection fails in handshak with javax.net.ssl.SSLException: Rec…

mbvreddy:
fixes https://github.com/eclipse/openj9/issues/3637
fixes https://github.com/eclipse/openj9/issues/3618

Handshake failure with OpenSSL GCM (enabled by default) when connecting to https. Reason for the failure is that NativeCrypto.GCMEncrypt() is using 128-bit cipher (EVP_aes_128_gcm()) to encrypt key of length 256 bits. Changing it to use EVP_aes_256_gcm() solves the issue.


charliegracie:
I pushed a commit with some suggested changes.  Feel free to squash it into your commit, or leave it ore remove it if you do not think it is correct.

mbvreddy:
@enasser switched to switch-case.

pshipton:
jenkins compile win

closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
```diff
@@ -423,9 +423,19 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_GCMEncrypt
         OpenSSL_add_all_algorithms();
         ERR_load_crypto_strings();
         first_time_gcm = 1;
-
-        evp_gcm_cipher = EVP_aes_128_gcm();
-    } 
+    }
+    
+    switch(keyLen) {
```
pshipton: It seems the switch should assert in the default case where the keyLen is not 16, 24, 32. I'm not sure what happens when NULL is passed to EVP_CipherInit_ex(). Although maybe it sets an easily debuggable error code.

PR #215, opened 2018-11-21
Add build version to native-test-libs

theresa-m:
Fixes: https://github.com/eclipse/openj9/issues/3584

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

reviewer @keithc-ca

### DIFF
--- a/closed/make/Main.gmk
+++ b/closed/make/Main.gmk
@@ -41,7 +41,11 @@ j9vm-build : openssl-build
 j9vm-compose-buildjvm : j9vm-build
 	+$(OPENJ9_MAKE) stage_openj9_build_jdk
 
-images :: test-image
+java-version-file: source-tips demos nashorn images-only test-image
+	# create file with build information in native libraries folder to identify which build the libraries belong to
+	$(JDK_IMAGE_DIR)/bin/java -version > $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt 2>&1
+
+images :: java-version-file
 
 test-image : openj9-test-image
 


make/Main.gmk
```diff
@@ -136,6 +136,9 @@ demos-only: start-make
 # Note: This double-colon rule is intentional, to support
 # custom make file integration.
 images:: source-tips demos nashorn images-only
+	# create file with build information in native libraries folder to identify which build the libraries belong to
+	$(JDK_IMAGE_DIR)/bin/java -version > $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt 2>&1
```
pshipton: I think this should be added to closed/make/Main.gmk

pshipton: fyi https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/114/files

closed/make/Main.gmk
```diff
@@ -38,8 +38,10 @@ openssl-build :
 j9vm-build : openssl-build
 	+$(OPENJ9_MAKE) OPENJDK_VERSION_NUMBER_FOUR_POSITIONS=$(VERSION_NUMBER_FOUR_POSITIONS) build-j9
 
-j9vm-compose-buildjvm : j9vm-build
+j9vm-compose-buildjvm : images-only
```
keithc-ca: I think this needs to depend upon `exploded-images` and not just `images-only`, otherwise `java` may not run.

theresa-m: `exploded-image` does not exist in Java 8, only in newer versions

keithc-ca: That's right, but my point is that this must depend on something that will guarantee that `java` is executable; `*-only` targets don't qualify. See the top-level Makefile:

```
$(info .  make <component>       # Build <component> and everything it depends on. )
$(info .  make <component>-only  # Build <component> only, without dependencies. This)
$(info .                         # is faster but can result in incorrect build results!)
```

theresa-m: okay. maybe its better to move this back to main/Main.gmk under the images target? I can't do `j9vm-compose-buildjvm : images` here because it will not refer to the wrong images target here https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/closed/make/Main.gmk#L44

keithc-ca: You should be able to make your changes in `closed/main/Main.gmk`.

keithc-ca:
The dependencies have been worked out in #269.

PR #217, opened 2018-11-26
Use -pagezero_size always on OSX

babsingh:
-pagezero_size 0x1000 can be used for both compressedrefs and
non-compressedrefs variants. This will allow executables to work with
both compressedrefs and non-compressedrefs OpenJ9 librarires.

### DIFF
--- a/jdk/make/closed/autoconf/custom-spec.gmk.in
+++ b/jdk/make/closed/autoconf/custom-spec.gmk.in
@@ -94,8 +94,7 @@ ifeq ($(OPENJDK_BUILD_OS), macosx)
   # variable is defined to support dependencies where the compiler option
   # is not applied.
   export MACOSX_DEPLOYMENT_TARGET := @MACOSX_VERSION_MIN@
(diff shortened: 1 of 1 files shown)


babsingh:
fyi - @DanHeidinga

DanHeidinga:
@babsingh I'm not sure I follow the rationale for this change.  Is it solving a problem?  Additional details on *why* this change is required would help

babsingh:
First, it will allow us to access memory below 4GB in `non-compressedrefs`.

If I use `compressedrefs` shared libraries in a `non-compressedrefs` SDK, then the executables such as `java`, `javap` ... would fail with `JVMJ9VM015W Initialization error for library j9gc29(2): Failed to instantiate compressed references metadata; 200M requested`. 

Second, this change will provide a generic front-end where executables such as `java`, `javap` ... can be used with both `compressedrefs` and `non-compressedrefs` shared libraries.

DanHeidinga:
issue: eclipse/openj9#3861

DanHeidinga:
No PR builds for osx mac jdk8 yet.  Merging as is

PR #218, opened 2018-11-28
Restoring ResourceBundle to the OpenJDK version

adamfarley:
The latest ResourceBundle.java in a tagged version of OpenJDK8
contains a fix that renders our patch here superfluous.

Replacing this file with the version from OpenJDK8.

Note that this file will be different from master until we update
master again, just prior to the next release.

### DIFF
--- a/jdk/src/share/classes/java/util/ResourceBundle.java
+++ b/jdk/src/share/classes/java/util/ResourceBundle.java
@@ -24,11 +24,9 @@
  */
 
 /*
- * ===========================================================================
- * (c) Copyright Taligent, Inc. 1996, 1997 All Rights Reserved
- * ===========================================================================
- * (c) Copyright IBM Corp. 1996, 2018 All Rights Reserved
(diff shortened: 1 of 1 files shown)


lumpfish:
@groeges - this change (reverting to the openjdk version of the file) causes the copyright checker to think there's an issue.  Is this is a situation we need the checker to cope with, or can you think of a better way of achieving the reversion?

groeges:
@lumpfish This will still be using the old copyrightCheck as the new one is not live yet. The old one just checked to see if the year in the IBM copyright was the current year. 

If we are reverting a file then I think even the new copyrightCheck will still flag this as an issue as we will be updating a file in the PR but it will not have an IBM copyright. 

This will just have to be identified in such PRs and an exception taken to merge the PR with the copyrightCheck failure.

adamfarley:
This PR is not needed, as the upstream project has been updated with the same fix, and we've integrated that fix.

PR #227, opened 2018-12-19
Dynamic loading of crypto library and support for OpenSSL 1.0.2

enasser:
In this changeset, the following changes are added:
    - Support of dynamic loading of OpenSSL crypto library
    - Support for 1.0.2, 1.1.0 and 1.1.1 OpenSSL releases

### DIFF
--- a/closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
+++ b/closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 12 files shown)


enasser:
@pshipton I have made the IBM Copyright with year as 2019 and the Copyright Check failed. Whether the tool will be corrected or I need to revert to 2018?

enasser:
I have made the changes for all platforms. Completed building and tested on xa64. Now building on other platforms like windows, AIX, zLinux and ppcle.

pshipton:
@enasser re the copyrights, I think you missed updating some file to 2019.
```
08:49:02 The following files were modified and have incorrect copyrights
[Pipeline] echo
08:49:02 closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 closed/get_openssl_source.sh
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 closed/make/mapfiles/libjncrypto/mapfile-vers
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 closed/openssl.gmk
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 jdk/make/closed/autoconf/custom-hook.m4
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 jdk/make/closed/autoconf/custom-spec.gmk.in
08:49:02 	IBM Copyright date appears to be incorrect
[Pipeline] echo
08:49:02 jdk/make/lib/SecurityLibraries.gmk
08:49:02 	IBM Copyright date appears to be incorrect
```

SueChaplain:
@enasser - probably should have been on my radar from the weekly calls, but it wasn't clear to me that this work was being done for 0.12.0. Presumably it requires documentation? If so, please raise an item at https://github.com/eclipse/openj9-docs/issues/new?template=new-documentation-change.md

enasser:
@SueChaplain Yes. It is targeted for 0.12. I will create an issue for documentation changes.

enasser:
@andrew-m-leonard @pshipton Please review and merge.

enasser:
@andrew-m-leonard  I have fixed a compile error for macos platform and pushed as a separate commit. Will squash it once you complete the review and ready for merge.

pshipton:
@enasser note the signed-off-by in the mac commit is incorrect (missing `<` `>`)

pshipton:
jenkins compile xlinux,zlinux,plinux,osx

pshipton:
jenkins compile zlinux

pshipton:
@jdekonin I can't start zlinux compile testing.

pshipton:
jenkins test sanity zlinux

pshipton:
jenkins compile win

pshipton:
jenkins compile xlinux depends eclipse/openj9#4291

pshipton:
osx compiled successfully with eclipse/openj9#4291

pshipton:
jenkins test sanity zlinux depends eclipse/openj9#4291

pshipton:
@enasser I think this is ready to go once the changes are squashed and the signed-off-by check is passing.

Does this work for Windows, is it possible to use `--with-openssl=system` on Windows?

pshipton:
jenkins compile aix depends eclipse/openj9#4291

pshipton:
@enasser actually, if this has the same compile problem with gcc 7.3 as jdk11 https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/90#issuecomment-454186839, this needs to be fixed before merge.

pshipton:
jenkins compile plinux depends eclipse/openj9#4291

enasser:
@pshipton  regarding your question:
> Does this work for Windows, is it possible to use --with-openssl=system on Windows?

Yes. Now we need only the header file to compile. Hence, --with-openssl=system works for all platforms including windows if openssl is package installed (otherwise header files are not included in the binary).

andrew-m-leonard:
@enasser Looks good, but i'll leave it to @pshipton to merge as i'm not on top of all the comments and changes above...

enasser:
@pshipton Regarding your comment

> @enasser actually, if this has the same compile problem with gcc 7.3 as jdk11 ibmruntimes/openj9-openjdk-jdk11#90 (comment), this needs to be fixed before merge.

I have responded to this comment in ibmruntimes/openj9-openjdk-jdk11. We can Use --disable-warnings-as-errors compile option to avoid the failure. The code is same for Java 8 and Java 11. When Java 8 compiler is upgraded to gcc 7.3, we can expect the same error for Java 8 as well.  

We need to see how to resolve this compiler warning by cleanup the code. I think it can be considered later and we can go ahead with this code now considering the shortage of time.. I have built and tested the code to ensure it's working as expected.

pshipton:
Seems my builds with `depends eclipse/openj9#4291` didn't pick up the 4291 changes. I'll need to re-test this.

pshipton:
plinux testing in eclipse/openj9#4291 results in
`08:42:25 configure: error: Unable to find openssl 1.1.0(and above) installed on System. Please use other options for '--with-openssl'`

@enasser is this a problem with the error message? Is 1.1.x+ required, or is 1.0.x+ sufficient?

pshipton:
I suppose this could be delivered as is. Compiling with gcc 7 doesn't cause any errors on jdk8, I suspect warnings as errors is already enabled somehow for jdk8. However since jdk11 can't be delivered as-is, and this is the same code, I won't merge it now. I've also only been able to verify xlinux working with `--with-openssl=system` so far, as openssl is not yet installed on the other OSes https://github.com/eclipse/openj9/issues/4293. In addition, we are added a test https://github.com/eclipse/openj9/pull/4300 to confirm the jncrypto can be successfully loaded, which will be helpful to verify this change.

enasser:
@pshipton Made the changes similar to java 11 to resolve the compile errors, squash and push. Please review and merge.

pshipton:
jenkins compile zlinux,win

pshipton:
jenkins compile zlinux

pshipton:
jenkins compile plinux

pshipton:
jenkins test sanity xlinux

closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
```diff
@@ -56,15 +121,159 @@ static void handleErrors(void) {
...
+     // Determine the version of OpenSSL.
+    OSSL_version_t* OSSL_version;
+    const char * openssl_version;
+    int ossl_ver;
+
+
+    // Load OpenSSL Crypto library
+    handle = load_crypto_library();
+    if (handle == NULL) {
+        fprintf(stderr, "FAILED TO LOAD OPENSSL CRYPTO LIBRARY\n");
```
pshipton: I don't think the code should be printing to the console when it fails to load the crypto library, it should silently fail.

pshipton:
I'm concerned about the code printing the console when the crypto library can't be loaded, but I'm going to merge it anyway and this can be resolved afterwards.

pshipton:
In combination with #239, this change seems to be causing build problems and blocking at least Windows builds. I'm going to revert it. Note this error is also seen in the equivalent jdk11 change https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/90.
`22:01:32 make[2]: *** No rule to make target '/cygdrive/c/OpenSSL-1.1.1-x86_64/libeay32.dll', needed by '/cygdrive/c/Users/jenkins/workspace/PullRequest-Compile-JDK8-win_x86-64_cmprssptrs-OpenJ9/build/windows-x86_64-normal-server-release/jdk/bin/libeay32.dll'.`

enasser:
@pshipton  The make file failure can occur when libcrypto-1_1-x64.dll  does not exist at /cygdrive/c/OpenSSL-1.1.1-x86_64. I will modify the code to avoid the failure and just warn when the bundling library is not available in the expected path.

enasser:
@pshipton As you explained, in the failing machine, libcrypto-1_1-x64.dll is available  at /cygdrive/c/OpenSSL-1.1.1-x86_64/bin. Will include this path as well for the library search path.

PR #233, opened 2019-01-11
Ignore complaints by 'strip' applied to already stripped objects

keithc-ca:
On AIX, the VM libraries have already been stripped and strip will fail when invoked a second time: ignore such failures.


keithc-ca:
This is likely unnecessary once https://github.com/eclipse/openj9/pull/4276 is merged.

keithc-ca:
Closing: This will be unnecessary once https://github.com/eclipse/openj9/pull/4286 is merged.

PR #237, opened 2019-01-14
Remove special cases for AIX

keithc-ca:
`export LDR_CNTRL=MAXDATA=0x40000000` conflicts with the way AIX machines have been set up (and 1GB seems to be insufficient lately).


keithc-ca:
Other than the copyright addition (which could be removed) this removes any differences from the file from openjdk.

pshipton:
PR testing for this is running in #4286

keithc-ca:
This appears to have fixed eclipse/openj9#4120.

pshipton:
Ignoring the copyright failure since this change reverts the file to remove all IBM changes.

pshipton:
@sxa555 this should fix `virtual memory exhausted` for AIX jdk8, although there will be another failure until https://github.com/eclipse/openj9/pull/4286 is merged as well.

PR #238, opened 2019-01-16
Collapse rules to copy LICENSE

keithc-ca:
Reduce repetition of command to copy license files.


keithc-ca:
The jdk8 compile job completed successfully, despite what this page says.

PR #239, opened 2019-01-18
Re-run common/autoconf/autogen.sh

keithc-ca:
Required to incorporate changes in #227.


pshipton:
In combination with #227, there are build problems which are blocking at least Windows builds. I'm going to revert it.

PR #244, opened 2019-01-25
Allow loading of crypto library on Mac OSX

enasser:
The issue with mac OSX is that libjncrypto.dylib is linked to
/usr/local/lib/libcrypto.1.1.dylib instead of linking with rpath
i.e @rpath/libcrypto.1.1.dylib.
This is because libcrypto.1.1.dylib by default is built with
/usr/local/lib as the path instead of rpath.
To fix the issue, updated the path entry of libcrypto.1.1.dylib
with @rpath using install_name_tool.

### DIFF
--- a/jdk/make/CopyFiles.gmk
+++ b/jdk/make/CopyFiles.gmk
@@ -23,7 +23,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
 # ===========================================================================
 #
 
@@ -281,6 +281,14 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
     OPENSSL_TARGET_LIB = $(JDK_OUTPUTDIR)/bin/$(OPENSSL_LIB_NAME)
   else
     OPENSSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_LIB_NAME)
+  endif
+
+  # On Mac OS, update the crypto library path to @rpath as the default is /usr/local/lib/
+  ifeq ($(OPENJDK_BUILD_OS), macosx)
+    .PHONY : RPATH_CRYPTO_LIB
+    COPY_FILES += RPATH_CRYPTO_LIB
+    RPATH_CRYPTO_LIB :
+	install_name_tool -id "@rpath/$(OPENSSL_LIB_NAME)" "$(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_LIB_NAME)"
   endif
 
   $(OPENSSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_LIB_NAME)
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -1006,7 +1006,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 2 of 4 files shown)


keithc-ca:
Jenkins compile osx,plinux,win jdk8

pshipton:
@enasser fyi the test for osx in https://github.com/eclipse/openj9/pull/4450 has failed. Keith is looking into why.

jdk/make/CopyFiles.gmk
```diff
@@ -283,6 +283,14 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
     OPENSSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_LIB_NAME)
   endif
 
+  # On Mac OS, update the crypto library path to @rpath as the default is /usr/local/lib/
+  ifeq ($(OPENJDK_BUILD_OS), macosx)
+    .PHONY : RPATH_CRYPTO_LIB
+    TARGETS += RPATH_CRYPTO_LIB
```
keithc-ca: I think this needs to be `COPY_FILES` instead of `TARGETS` so there's a dependency on `RPATH_CRYPTO_LIB`, otherwise the command below will not execute. See `jdk/make/CompileNativeLibraries.gmk`:
> all: $(COPY_FILES) $(BUILD_LIBRARIES)

keithc-ca: Confirmed in https://ci.eclipse.org/openj9/job/Build-JDK8-osx_x86-64_cmprssptrs/88/:
```
$ objdump -dylib-id -macho jre/lib/libcrypto.1.1.dylib
jre/lib/libcrypto.1.1.dylib:
/usr/local/lib/libcrypto.1.1.dylib
```

keithc-ca: Even if I run `install_name_tool` manually, it doesn't fix the problem because `libjncrypto.dylib` was created before the right id is installed in `libcrypto.dylib`. I think the right time to run `install_name_tool` is during configuration - after the file is copied to `${LOCAL_CRYPTO}`.

keithc-ca: My previous comment applies when configured using `--with-openssl={path}`; when configure decides `BUILD_OPENSSL=yes`, then `install_name_tool` should be invoked within `closed/openssl.gmk`.

keithc-ca: I had been assuming the changes to `jdk/make/closed/autoconf/custom-hook.m4` were effective, but they won't be until you run `bash common/autoconf/autogen.sh` and push the new generated files to your branch.

enasser: Thank you @keithc-ca for your comments. As you identified, I missed to update generated-configure.sh (because of the influence of working with openjdk11 as it's automatically generated while building :-)). Sorry for the inconvenience.
@pshipton I have updated generated-configure.sh, squashed and pushed. Please resubmit the build jobs.

pshipton:
Running run https://github.com/eclipse/openj9/pull/4450
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-osx_x86-64_cmprssptrs-OpenJ9/72/
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-win_x86-64_cmprssptrs-OpenJ9/489/
https://ci.eclipse.org/openj9/job/PullRequest-Sanity-JDK8-linux_x86-64_cmprssptrs-OpenJ9/639/

pshipton:
@enasser it is still not working for osx. Please try the other changes that Keith suggested.

pshipton:
New testing started in https://github.com/eclipse/openj9/pull/4450

pshipton:
@enasser the testing has passed. Can you please cherry pick this change and create a PR against the openj9-0.12 branch.

jdk/make/CopyFiles.gmk
keithc-ca: I'm not convinced that, as merged, we are guaranteed that `install_name_tool` will be used to update `libcrypto.dylib` *before* `libjncrypto.dylib` is created and so we can't be sure it will have the proper reference to `libcrypto.dylib`. Unfortunately, I also don't have a Mac that is capable of building jdk8.

PR #245, opened 2019-01-28
Allow loading of crypto library on Mac OSX

enasser:
The issue with mac OSX is that libjncrypto.dylib is linked to
/usr/local/lib/libcrypto.1.1.dylib instead of linking with rpath
i.e @rpath/libcrypto.1.1.dylib.
This is because libcrypto.1.1.dylib by default is built with
/usr/local/lib as the path instead of rpath.
To fix the issue, updated the path entry of libcrypto.1.1.dylib
with @rpath using install_name_tool.

### DIFF
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -1006,7 +1006,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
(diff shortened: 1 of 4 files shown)


enasser:
@pshipton I have created the PR for openj9-0.12 branch as you suggested. Please review and merge.

pshipton:
Thanks Nasser.

PR #260, opened 2019-02-01
Turn off native digest

jerrylui803:
This change will disable OpenSSL based accelerated Digest, it cannot be re-enabled with the command line option. This was done due to a segfault under some conditions when the Digest is invoked.
https://github.com/eclipse/openj9/issues/4530

Same as https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/259 for release branch


pshipton:
Obsoleted by https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/264

PR #261, opened 2019-02-01
Revert "Turn off native digest"

pshipton:
Reverts ibmruntimes/openj9-openjdk-jdk8#259

There is nothing wrong with the change, but it is causing the OpenJ9 OpenSSL_0  and cmdLineTester_CryptoTest_0 testing to fail. The tests need to be fixed. Reverting until the fix is ready.


pshipton:
Reverting this revert in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/263

PR #263, opened 2019-02-01
Turn off native digest

pshipton:
Reverts ibmruntimes/openj9-openjdk-jdk8#261 to turn off native digest as per #259


pshipton:
Tested in https://github.com/eclipse/openj9/pull/4555

PR #266, opened 2019-02-08
Exclude _openj9 tag for version search

andrew-m-leonard:

### DIFF
--- a/jdk/make/closed/autoconf/custom-hook.m4
+++ b/jdk/make/closed/autoconf/custom-hook.m4
@@ -255,10 +255,11 @@ AC_DEFUN_ONCE([OPENJDK_VERSION_DETAILS],
   AC_SUBST(JDK_FIX_VERSION)
 
   OPENJDK_SHA=`git -C $SRC_ROOT rev-parse --short HEAD`
-  LAST_TAGGED_SHA=`git -C $SRC_ROOT rev-list --tags="jdk8u*" --topo-order --max-count=1 2>/dev/null`
+  # Find first rev tagged by jdk8u* but not containing "_openj9"
+  LAST_TAGGED_SHA=`git -C $SRC_ROOT rev-list --exclude="*_openj9*" --tags="jdk8u*" --topo-order --max-count=1 2>/dev/null`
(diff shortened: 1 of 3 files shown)


andrew-m-leonard:
jenkins compile xlinux

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -255,10 +255,11 @@ AC_DEFUN_ONCE([OPENJDK_VERSION_DETAILS],
   AC_SUBST(JDK_FIX_VERSION)
 
   OPENJDK_SHA=`git -C $SRC_ROOT rev-parse --short HEAD`
-  LAST_TAGGED_SHA=`git -C $SRC_ROOT rev-list --tags="jdk8u*" --topo-order --max-count=1 2>/dev/null`
+  # Find first rev tagged by jdk8u* but not containing "_openj9"
+  LAST_TAGGED_SHA=`git -C $SRC_ROOT rev-list --exclude="*_openj9*" --tags="jdk8u*" --topo-order --max-count=1 2>/dev/null`
   if test "x$LAST_TAGGED_SHA" != x ; then
     # Choose the latest tag when there is more than one for the same SHA.
-    OPENJDK_TAG=`git -C $TOPDIR tag --points-at "$LAST_TAGGED_SHA" | grep '+' | sort -V | tail -1`
+    OPENJDK_TAG=`git -C $TOPDIR tag --points-at "$LAST_TAGGED_SHA" | grep '-' | sort -V | tail -1`
```
lumpfish: Is the grep '-' assuming that the tags we are interested in are of the form jdk8u202-b08? So assuming there'll be no other tags with a hyphen and relying on 'b08' being collated ahead of '-ga'?  Just wondering whether a more restrictive check might be more obvious (jdk8u[\d]+-b[\d]+)?

andrew-m-leonard: the "+" was for jdk11 format tags where we want to not sort the "-ga" tag, so the grep filtered it out, as we're just interested in the jdk-11...+nn tags
For jdk8 format tags, the format is different it is "-bnn"
I can raise a new issue for that if we feel we need to fix that..

PR #269, opened 2019-02-12
Run 'java -version' if not cross compiling

keithc-ca:
Supercedes https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/215.

Issue ibmruntimes/openj9-openjdk-jdk11#133


pshipton:
jenkins compile zlinux

pshipton:
Reverting as I think this is causing test binaries like pltest to be missing when running functionality tests.

PR #271, opened 2019-02-13
Run 'java -version' if not cross compiling

keithc-ca:
Replays #269, adding the dependency: `all : test-image`


pshipton:
jenkins test sanity zlinux

keithc-ca:
This also corrects the target name (`test-image-openj9`) used by [1] on z/OS.

[1] https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/common/build#L270

PR #285, opened 2019-04-12
Improve git tag resolution

keithc-ca:
* now works with a shallow clone
* also works faster


pshipton:
jenkins compile zlinux

pshipton:
jenkins compile plinux

pshipton:
jenkins compile aix,win,osx

Issue #290, opened 2019-04-25
Bash version check not working correctly in all cases


liketechnik:
The script closed/get_j9_source.sh checks that the bash version in use it at least 4.0. The relevant line (https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/01c0db5569ce245ee610d30515f857b2cd984f4e/closed/get_j9_source.sh#L44) checks that the bash version string (first line of `bash --version`) contains 'version 4.'. 
The problem is that it checks for a lowercase 'v', which only works for the English output (LANG=en_EN). For example in the German localization the output has a capital 'V' so the check incorrectly treats the bash version as too old.
I think this problem could be solved by leaving out the 'v', e. g. checking only for containing 'ersion 4.' which would be the same for all localizations.

DanHeidinga:
@liketechnik I've opened a PR that should address this issue - #293 - using a different approach.  If you able to validate it works in your environment I'd appreciate it.

liketechnik:
The check kind of passes, but I get the following error:
```
closed/get_j9_source.sh: line 43: [: “04”: integer expression expected
```

pshipton:
@keithc-ca

keithc-ca:
I'll have another look at this.
@liketechnik : What does `bash --version` say for the version that gives you grief?

keithc-ca:
I think the problem is that file uses the wrong double-quote `“` instead of `"`.

liketechnik:
It's `GNU bash, version 4.4.23(1)-release (x86_64-pc-linux-gnu)`

PR #293, opened 2019-04-26
Make bash version check more inclusive

DanHeidinga:
close: #290

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

### DIFF
--- a/closed/get_j9_source.sh
+++ b/closed/get_j9_source.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2018 All Rights Reserved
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Jenkins compile plinux,zlinux,osx jdk8

DanHeidinga:
Jenkins compile win jdk8

DanHeidinga:
Note, this fix is already used in the  jdk11 & 12 repos.  I haven't confirmed the jdk repo but would be surprised if it wasn't already the case

keithc-ca:
> I haven't confirmed the jdk repo but would be surprised if it wasn't already the case

The version check is now consistent across all repos.

PR #303, opened 2019-05-16
Add OpenSSL related settings to CUSTOM_COMPILER_ENV_VARS

a7ehuo:
`JITServer` needs to be built with `libssl`. By default, it
is built against the `OpenSSL` installed on the build machine.

This change passes the `OpenSSL` build flags from OpenJDK to
JIT build environment.
- Add `OPENSSL_DIR`, `OPENSSL_CFLAGS`, `OPENSSL_LIBS`,
  and `OPENSSL_BUNDLE_LIB_PATH` to `CUSTOM_COMPILER_ENV_VARS`.
- Bundle `libssl` if `--with-openssl=fetched --enable-openssl-bundling`
  is set.

The JIT build change is addressed in: 
eclipse/openj9#5789, eclipse/openj9#7846

Examples on the supported flags:
```
--enable-jitserver --with-openssl=fetched
--enable-jitserver --with-openssl=fetched --enable-openssl-bundling
--enable-jitserver --with-openssl=system
```
Issue: eclipse/openj9#5720

Signed-off-by: Annabelle Huo <Annabelle.Huo@ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -408,11 +408,6 @@ endif
 ifneq (,$(OPENJ9_DEVELOPER_DIR))
   CUSTOM_COMPILER_ENV_VARS += DEVELOPER_DIR="$(OPENJ9_DEVELOPER_DIR)"
 endif
-ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
-  ifneq (true,$(OPENJ9_ENABLE_CMAKE))
-    CUSTOM_COMPILER_ENV_VARS += JITSERVER_SUPPORT=1
-  endif
-endif
 
 generate-j9-version-headers :
 	@$(ECHO) Ensuring version information is up-to-date
@@ -470,9 +465,21 @@ ifneq (,$(CCACHE))
 endif # CCACHE
 ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
   CMAKE_ARGS += -DJITSERVER_SUPPORT=ON
+
+  ifneq (,$(OPENSSL_CFLAGS))
+    CMAKE_ARGS += -DOPENSSL_CFLAGS="$(OPENSSL_CFLAGS)"
+  endif
+
+  ifneq (,$(OPENSSL_DIR))
+    CMAKE_ARGS += -DOPENSSL_DIR="$(OPENSSL_DIR)"
+  endif
+
+  ifneq (,$(OPENSSL_BUNDLE_LIB_PATH))
+    CMAKE_ARGS += -DOPENSSL_BUNDLE_LIB_PATH="$(OPENSSL_BUNDLE_LIB_PATH)"
+  endif
 else
   CMAKE_ARGS += -DJITSERVER_SUPPORT=OFF
-endif
+endif # OPENJ9_ENABLE_JITSERVER
 
   CMAKE_ARGS += $(EXTRA_CMAKE_ARGS)
 
@@ -484,6 +491,22 @@ $(OUTPUT_ROOT)/vm/cmake.stamp :
 run-preprocessors-j9 : $(OUTPUT_ROOT)/vm/cmake.stamp
 
 else # OPENJ9_ENABLE_CMAKE
+
+ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+  CUSTOM_COMPILER_ENV_VARS += JITSERVER_SUPPORT=1
+
+  ifneq (,$(OPENSSL_CFLAGS))
+    CUSTOM_COMPILER_ENV_VARS += OPENSSL_CFLAGS="$(OPENSSL_CFLAGS)"
+  endif
+
+  ifneq (,$(OPENSSL_DIR))
+    CUSTOM_COMPILER_ENV_VARS += OPENSSL_DIR="$(OPENSSL_DIR)"
+  endif
+
+  ifneq (,$(OPENSSL_BUNDLE_LIB_PATH))
+    CUSTOM_COMPILER_ENV_VARS += OPENSSL_BUNDLE_LIB_PATH="$(OPENSSL_BUNDLE_LIB_PATH)"
+  endif
+endif # OPENJ9_ENABLE_JITSERVER
 
 run-preprocessors-j9 : stage-j9
 	@$(ECHO) Running OpenJ9 preprocessors with OPENJ9_BUILDSPEC: $(OPENJ9_BUILDSPEC)
--- a/jdk/make/CopyFiles.gmk
+++ b/jdk/make/CopyFiles.gmk
@@ -323,7 +323,30 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
 
     COPY_FILES += $(OPENSSL_TARGET_LIB)
   endif
-endif
+
+  ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+    # Copy OpenSSL SSL Lbrary
+    # To bundle first search for libssl 1.1.x library, if not found, search for 1.0.x
+    ifeq ($(OPENJDK_TARGET_OS), linux)
+      OPENSSL_SSL_LIB_NAME = libssl.so.1.1
+      ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
+        OPENSSL_SSL_LIB_NAME = libssl.so.1.0.0
+        ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
+          OPENSSL_SSL_LIB_NAME =
+        endif
+      endif
+
+      ifneq ($(OPENSSL_SSL_LIB_NAME), )
+        OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)
+
+        $(OPENSSL_SSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
+			$(call install-file)
+
+        COPY_FILES += $(OPENSSL_SSL_TARGET_LIB)
+      endif # OPENSSL_SSL_LIB_NAME
+    endif # OPENJDK_TARGET_OS
+  endif # OPENJ9_ENABLE_JITSERVER
+endif # OPENSSL_BUNDLE_LIB_PATH
 
 ##########################################################################################
 
--- a/jdk/make/closed/autoconf/custom-hook.m4
+++ b/jdk/make/closed/autoconf/custom-hook.m4
@@ -702,6 +702,20 @@ AC_DEFUN([CONFIGURE_OPENSSL],
 
     AC_MSG_CHECKING([if we should bundle openssl])
     AC_MSG_RESULT([$BUNDLE_OPENSSL])
+
+    if test "x$OPENJ9_ENABLE_JITSERVER" = xtrue ; then
+      if test "x$OPENJDK_TARGET_OS" = xlinux ; then
+        if test "x$OPENSSL_DIR" != x ; then
+          AC_MSG_CHECKING([if the required OPENSSL API exists for JITServer in $OPENSSL_DIR])
+          if $GREP -q -w SSL_CTX_set_ecdh_auto "$OPENSSL_DIR/include/openssl/ssl.h" 2> /dev/null ; then
+            AC_MSG_RESULT([yes])
+          else
+            AC_MSG_RESULT([no])
+            AC_MSG_ERROR([SSL_CTX_set_ecdh_auto is required by JITServer])
+          fi
+        fi
+      fi
+    fi
   fi
 
   AC_SUBST(OPENSSL_BUNDLE_LIB_PATH)
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4397,7 +4397,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
-DATE_WHEN_GENERATED=1574278275
+DATE_WHEN_GENERATED=1576018331
 
 ###############################################################################
 #
--- a/jdk/make/closed/autoconf/generated-configure.sh
+++ b/jdk/make/closed/autoconf/generated-configure.sh
@@ -4576,7 +4576,7 @@ VS_SDK_PLATFORM_NAME_2017=
 
 
 # Do not change or remove the following line, it is needed for consistency checks:
-DATE_WHEN_GENERATED=1574278275
+DATE_WHEN_GENERATED=1576018331
 
 ###############################################################################
 #
@@ -55786,6 +55786,23 @@ $as_echo "yes" >&6; }
 $as_echo_n "checking if we should bundle openssl... " >&6; }
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: $BUNDLE_OPENSSL" >&5
 $as_echo "$BUNDLE_OPENSSL" >&6; }
+
+    if test "x$OPENJ9_ENABLE_JITSERVER" = xtrue ; then
+      if test "x$OPENJDK_TARGET_OS" = xlinux ; then
+        if test "x$OPENSSL_DIR" != x ; then
+          { $as_echo "$as_me:${as_lineno-$LINENO}: checking if the required OPENSSL API exists for JITServer in $OPENSSL_DIR" >&5
+$as_echo_n "checking if the required OPENSSL API exists for JITServer in $OPENSSL_DIR... " >&6; }
+          if $GREP -q -w SSL_CTX_set_ecdh_auto "$OPENSSL_DIR/include/openssl/ssl.h" 2> /dev/null ; then
+            { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes" >&5
+$as_echo "yes" >&6; }
+          else
+            { $as_echo "$as_me:${as_lineno-$LINENO}: result: no" >&5
+$as_echo "no" >&6; }
+            as_fn_error $? "SSL_CTX_set_ecdh_auto is required by JITServer" "$LINENO" 5
+          fi
+        fi
+      fi
+    fi
   fi
 
 


a7ehuo:
@ymanton @mpirvu For this change, I wonder if we should branch off `jitaas` first in `openj9-openjdk-jdk8` since we are not only exporting the required OpenSSL flags from openj9-openjdk-jdk8 but also bundle libssl which is not needed for the rest of JVM. However, it would also mean we need to maintain and rebase `jitaas` in openj9-openjdk-jdk8 as we do for OpenJ9 and OMR.

closed/OpenJ9.gmk
```diff
@@ -411,6 +411,18 @@ endif
...
+      CUSTOM_COMPILER_ENV_VARS += OPENSSL_CFLAGS="$(OPENSSL_CFLAGS)"
+    endif
+
+    ifneq (,$(OPENSSL_DIR))
+      CUSTOM_COMPILER_ENV_VARS += OPENSSL_DIR="$(OPENSSL_DIR)"
+    endif
+
+    ifneq (,$(OPENSSL_BUNDLE_LIB_PATH))
+      CUSTOM_COMPILER_ENV_VARS += OPENSSL_BUNDLE_LIB_PATH="$(OPENSSL_BUNDLE_LIB_PATH)"
+    endif
```
keithc-ca: I suggest this block should be near the equivalent cmake-related stuff below.

jdk/make/CopyFiles.gmk
```diff
@@ -323,6 +323,29 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
 
     COPY_FILES += $(OPENSSL_TARGET_LIB)
   endif
+
+  ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+    # Copy OpenSSL SSL Lbrary
+    # To bundle first search for libssl 1.1.x library, if not found, search for 1.0.x
+    ifeq ($(OPENJDK_TARGET_OS), linux)
+      OPENSSL_SSL_LIB_NAME = libssl.so.1.1
+      ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
```
keithc-ca: If openssl is not bundled, `OPENSSL_BUNDLE_LIB_PATH` will not be set: how will this work?

a7ehuo: In the case where openssl is not bundled and`OPENSSL_BUNDLE_LIB_PATH` is not set: 

1. If `--with-openssl=fetched` is specified, `OPENSSL_DIR ` is set. JITServer is built against the fetched OpenSSL under `OPENSSL_DIR`. This is the main use case since the PR Jenkins build specifies `--with-openssl=fetched`. 

2. If `--with-openssl=system` is specified or `--with-openssl` is not set at all, JITServer is built against the installed OpenSSL on the build system. Without this change, this is what we have now by default.

keithc-ca: It should be possible to add jitserver SSL support independent of whether the extra libraries are bundled or not, and independent of where the required libraries are on the build system.

a7ehuo: Similar reply as https://github.com/eclipse/openj9/pull/7846#discussion_r351011230

a7ehuo:
@keithc-ca All comments addressed. Ready for another review. Thanks. 
@mpirvu FYI.

jdk/make/CopyFiles.gmk
```diff
@@ -323,6 +323,29 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
...
+
+    ifneq ($(OPENSSL_SSL_LIB_NAME), )
+      OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)
+
+      $(OPENSSL_SSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
+		$(CP) $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME) $@
+
+      COPY_FILES += $(OPENSSL_SSL_TARGET_LIB)
+    endif
+  endif
```
keithc-ca: These changes should be applied to `closed/make/CopyFiles.gmk`.

a7ehuo:
@keithc-ca All comments addressed. Ready for another review. Thanks.

closed/OpenJ9.gmk
```diff
@@ -485,6 +492,22 @@ run-preprocessors-j9 : $(OUTPUT_ROOT)/vm/cmake.stamp
...
+  endif
+
+  ifneq (,$(OPENSSL_DIR))
+    CUSTOM_COMPILER_ENV_VARS += OPENSSL_DIR="$(OPENSSL_DIR)"
+  endif
+
+  ifneq (,$(OPENSSL_BUNDLE_LIB_PATH))
+    CUSTOM_COMPILER_ENV_VARS += OPENSSL_BUNDLE_LIB_PATH="$(OPENSSL_BUNDLE_LIB_PATH)"
+  endif
+endif
```
keithc-ca: Please add `# OPENJ9_ENABLE_JITSERVER` to the end of this line to make it easier to recognize the matching `ifeq`.

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -702,6 +702,21 @@ AC_DEFUN([CONFIGURE_OPENSSL],
...
+        if test "x$OPENSSL_CFLAGS" != x; then
+          CHECK_OPENSSL_HEADER=`$ECHO $OPENSSL_CFLAGS | $SED -e 's/\-I//'`
+          CHECK_OPENSSL_HEADER="${CHECK_OPENSSL_HEADER}/openssl/ssl.h"
+          AC_MSG_CHECKING([if the required OPENSSL API exists for JITServer in $CHECK_OPENSSL_HEADER])
+          if $GREP SSL_CTX_set_ecdh_auto "$CHECK_OPENSSL_HEADER" 2>&1 > /dev/null ; then
+            AC_MSG_RESULT([yes])
+          else
+            AC_MSG_ERROR([no, SSL_CTX_set_ecdh_auto is required by JITServer])
+          fi
+        fi
```
keithc-ca: * The check should occur even if `OPENSSL_CFLAGS` is empty (meaning no special compile options are necessary) and it's unreasonable to assume that `OPENSSL_CFLAGS` consists only of a single `-I` option: You should just use `$OPENSSL_DIR/include/openssl/ssl.h`.

* Both branches of the `if` should have `AC_MSG_RESULT`.

```
  AC_MSG_CHECKING([existence of OPENSSL API required by JITServer in openssl/ssl.h])
  if $GREP -qw SSL_CTX_set_ecdh_auto "$OPENSSL_DIR/include/openssl/ssl.h" 2> /dev/null ; then
    AC_MSG_RESULT([yes])
  else
    AC_MSG_RESULT([no])
    AC_MSG_ERROR([SSL_CTX_set_ecdh_auto is required by JITServer])
  fi
```
* Also notice the use of grep options `-q` and `-w`, and the change to stderr redirection.

closed/make/CopyFiles.gmk
```diff
@@ -37,6 +37,31 @@ ifeq ($(OPENJDK_TARGET_OS), windows)
...
+    ifneq ($(OPENSSL_SSL_LIB_NAME), )
+      OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)
+
+      $(OPENSSL_SSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
+		$(CP) $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME) $@
+
+      COPY_FILES += $(OPENSSL_SSL_TARGET_LIB)
+    endif # OPENSSL_SSL_LIB_NAME
+  endif # OPENJ9_ENABLE_JITSERVER
+endif # OPENSSL_BUNDLE_LIB_PATH
```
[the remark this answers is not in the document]
a7ehuo: [1] copies `libcrypto` not `libssl`.  This part of the code makes sure `libssl` is bundled. 

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/make/CopyFiles.gmk#L252-L326

jdk/make/closed/autoconf/custom-hook.m4
a7ehuo: * `OPENSSL_DIR` is not set when `--with-openssl=system` is specified but `OPENSSL_CFLAGS` will get set.
* Will add `AC_MSG_RESULT([no])`. 
* Will change it to `if $GREP -q -w SSL_CTX_set_ecdh_auto "$CHECK_OPENSSL_HEADER" ;`

closed/make/CopyFiles.gmk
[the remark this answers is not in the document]
keithc-ca: Ok, I missed that difference; I think the part existing code that copies `libcypto` should also copy `libssl` rather than have the copying in two places.

jdk/make/closed/autoconf/custom-hook.m4
keithc-ca: In the `--with-openssl=system` case, I would expect some variable to be set by `PKG_CHECK_MODULES(OPENSSL, ...)` that can be use to set `OPENSSL_DIR ` (or that variable should be used throughout instead of `OPENSSL_DIR`).

a7ehuo: [1] indicates `OPENSSL_CFLAGS` and `OPENSSL_LIB` will be set up by `PKG_CHECK_MODULES`. My test shows `cmake version 3.5.1` will set them up but `cmake version 3.10.2` will not with our current `jdk/make/closed/autoconf/custom-hook.m4`. Using `OPENSSL_CFLAGS` here just gives more cases to check this one corner case of `SSL_CTX_set_ecdh_auto`. 

[1]
https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/8514ba3391a835583387d5065fccd009c13379c6/jdk/make/closed/autoconf/custom-hook.m4#L629

keithc-ca: Both versions of cmake set `OPENSSL_INCLUDE_DIR` which seems like the right place to look for header files and the right name to use when `--with-openssl` is not `system`.

a7ehuo: Sorry, I mixed up the `autoconf PKG_CHECK_MODULES` with the` cmake pkg_check_modules`. We are using `autoconf PKG_CHECK_MODULES` in `custom-hook.m4`. 

`PKG_CHECK_MODULES` sets only two variables `<>_CFLAGS` and `<>_LIBS` [1]. 

There are also two scenarios to consider when setting `--with-openssl=system`. 

1. If OpenSSL is installed at the default system location such as `/usr`, `pkg-config` could strip out the flags. `<>_CFLAGS` and `<>_LIBS` will not be set by `PKG_CHECK_MODULES` [2]. 

Running bash configure as below will NOT set  `<>_CFLAGS` and `<>_LIBS`:
```
bash configure --with-freemarker-jar=/root/freemarker.jar --with-boot-jdk=/root/bootjdk8 --enable-jitserver --with-openssl=system` 
```

Running bash configure with `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` and  `PKG_CONFIG_ALLOW_SYSTEM_LIBS ` will have both `<>_CFLAGS` and `<>_LIBS` set:
```
bash configure --with-freemarker-jar=/root/freemarker.jar --with-boot-jdk=/root/bootjdk8 --enable-jitserver --with-openssl=system PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 PKG_CONFIG_ALLOW_SYSTEM_LIBS=1``` 
...
OPENSSL_CFLAGS=-I/usr/include
OPENSSL_LIBS=-L/usr/lib/x86_64-linux-gnu -lssl -lcrypto
``` 

2. If OpenSSL is installed at non default locations such as `/usr/local`, `PKG_CHECK_MODULES` will set `<>_CFLAGS` and `<>_LIBS`
```
bash configure --with-freemarker-jar=/root/freemarker.jar --with-boot-jdk=/root/bootjdk8 --enable-jitserver --with-openssl=system
...
OPENSSL_CFLAGS=-I/usr/local/include
OPENSSL_LIBS=-L/usr/local/lib -lssl -lcrypto
```
---------------------------------------------------------------------------
[1] https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/8514ba3391a835583387d5065fccd009c13379c6/common/autoconf/build-aux/pkg.m4#L128-L131
[2] https://linux.die.net/man/1/pkg-config

jdk/make/CopyFiles.gmk
```diff
@@ -323,7 +323,30 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
...
+    endif # OPENJDK_TARGET_OS
+
+    ifneq ($(OPENSSL_SSL_LIB_NAME), )
+      OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)
+
+      $(OPENSSL_SSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
+		$(CP) $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME) $@
+
+      COPY_FILES += $(OPENSSL_SSL_TARGET_LIB)
+    endif # OPENSSL_SSL_LIB_NAME
```
keithc-ca: This should be within the `if` above where `OPENSSL_SSL_LIB_NAME` is set.

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -702,6 +702,20 @@ AC_DEFUN([CONFIGURE_OPENSSL],
 
     AC_MSG_CHECKING([if we should bundle openssl])
     AC_MSG_RESULT([$BUNDLE_OPENSSL])
+
+    if test "x$OPENJ9_ENABLE_JITSERVER" = xtrue ; then
+      if test "x$OPENJDK_TARGET_OS" = xlinux ; then
+        if test "x$OPENSSL_DIR" != x; then
```
keithc-ca: nit: spaces both sides of `;` for consistency

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -702,6 +702,20 @@ AC_DEFUN([CONFIGURE_OPENSSL],
...
+          AC_MSG_CHECKING([if the required OPENSSL API exists for JITServer in $OPENSSL_DIR])
+          if $GREP -q -w SSL_CTX_set_ecdh_auto "$OPENSSL_DIR/include/openssl/ssl.h" ; then
```
keithc-ca: This should probably redirect stderr to avoid unwanted noise if the file does not exist.

jdk/make/CopyFiles.gmk
a7ehuo: Do you mean change it to the following? 

```
  ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
    # Copy OpenSSL SSL Lbrary
    # To bundle first search for libssl 1.1.x library, if not found, search for 1.0.x
    ifeq ($(OPENJDK_TARGET_OS), linux)
      OPENSSL_SSL_LIB_NAME = libssl.so.1.1
      ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
        OPENSSL_SSL_LIB_NAME = libssl.so.1.0.0
        ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
          OPENSSL_SSL_LIB_NAME =
        endif
      endif

      ifneq ($(OPENSSL_SSL_LIB_NAME), )
        OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)

        $(OPENSSL_SSL_TARGET_LIB): $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
		  $(CP) $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME) $@

        COPY_FILES += $(OPENSSL_SSL_TARGET_LIB)
      endif # OPENSSL_SSL_LIB_NAME
    endif # OPENJDK_TARGET_OS
  endif # OPENJ9_ENABLE_JITSERVER
```

keithc-ca: Yes. Structured this way, you don't have to worry about any other possible source for `OPENSSL_SSL_LIB_NAME` (like the environment).

closed/OpenJ9.gmk
```diff
@@ -470,9 +465,17 @@ ifneq (,$(CCACHE))
 endif # CCACHE
 ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
   CMAKE_ARGS += -DJITSERVER_SUPPORT=ON
+
+  ifneq (,$(OPENSSL_DIR))
+    CMAKE_ARGS += -DOPENSSL_DIR="$(OPENSSL_DIR)"
+  endif
```
keithc-ca: Why is `OPENSSL_CFLAGS` not propagated to the VM build here, like it is below?

jdk/make/CopyFiles.gmk
```diff
@@ -323,7 +323,30 @@ ifneq ($(OPENSSL_BUNDLE_LIB_PATH), )
...
+        ifeq ("$(wildcard $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME))", "")
+          OPENSSL_SSL_LIB_NAME =
+        endif
+      endif
+
+      ifneq ($(OPENSSL_SSL_LIB_NAME), )
+        OPENSSL_SSL_TARGET_LIB = $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)/$(OPENSSL_SSL_LIB_NAME)
+
+        $(OPENSSL_SSL_TARGET_LIB) : $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME)
+			$(CP) $(OPENSSL_BUNDLE_LIB_PATH)/$(OPENSSL_SSL_LIB_NAME) $@
```
keithc-ca: This should say `$(call install-file)` instead of using `$(CP)`.

closed/OpenJ9.gmk
a7ehuo: cmake calls `include_directories()` which would have the same effect as adding OPENSSL_CFLAGS to J9_CXXFLAGS and J9_CFLAGS.
https://github.com/eclipse/openj9/pull/7846#discussion_r356862010

keithc-ca: We should not assume that `OPENSSL_CFLAGS` only contains `-I` options.

a7ehuo: As per our offline discussion, I shouldn't assume `OPENSSL_CFLAGS` would contain only include path information. Will set it up in `CMAKE_ARGS` and pass it to `J9_CXXFLAGS` and `J9_CFLAGS`.

keithc-ca:
Jenkins compile plinux,plinuxjit jdk8

mpirvu:
FYI: Testing plinux and plinuxjit at the same time may now work at the moment: https://github.com/eclipse/openj9/pull/7860#issuecomment-561917396
The fix is not merged yet: https://github.com/eclipse/openj9/pull/7860#issuecomment-561917396

keithc-ca:
Even if there's a real problem with the plinuxjit build, I don't think it should hold this up.

a7ehuo:
The failure is because `JITServer` PPC build is not working yet. It requires https://github.com/eclipse/openj9/pull/7846, which adds the following to PPC.

@keithc-ca Could the dependency to https://github.com/eclipse/openj9/pull/7846 be added to `plinuxjit` PR build or maybe we could test `xlinux` and `xlinuxjit` instead? 

```
ifeq ($(HOST_ARCH),p)
...
    ifneq ($(JITSERVER_SUPPORT),)
        CX_DEFINES+=JITSERVER_SUPPORT
    endif
endif
```
```
12:07:33  /home/jenkins/workspace/Build_JDK8_ppc64le_linux_jit_Personal/build/linux-ppc64le-normal-server-release/vm/compiler/../compiler/env/j9methodServer.cpp:39:37: error: ‘class TR::CompilationInfoPerThread’ has no member named ‘getClientData’; did you mean ‘getMetadata’?
12:07:33      auto &classMap = threadCompInfo->getClientData()->getROMClassMap();
12:07:33                                       ^~~~~~~~~~~~~
12:07:33                                       getMetadata
```

a7ehuo:
The failure is not related to this change.

keithc-ca:
I'll do testing of plinuxjit in eclipse/openj9#7846 after this is merged.

PR #307, opened 2019-05-27
AArch64: Allow aarch64 as the target processor

knn-k:
This commit adds support for AArch64 in configuring the build.

Signed-off-by: knn-k <konno@jp.ibm.com>

### DIFF
--- a/jdk/make/closed/autoconf/generated-configure.sh
+++ b/jdk/make/closed/autoconf/generated-configure.sh
@@ -15196,6 +15196,9 @@ fi
     powerpc64)
       OPENJ9_CPU=ppc-64
       ;;
+    aarch64)
+      OPENJ9_CPU=aarch64
+      ;;
     *)
       as_fn_error $? "unsupported OpenJ9 cpu $build_cpu" "$LINENO" 5
       ;;
@@ -15243,6 +15246,8 @@ fi
     OPENJ9_PLATFORM_CODE=xz64
   elif test "x$OPENJ9_CPU" = xppc-64 ; then
     OPENJ9_PLATFORM_CODE=ap64
+  elif test "x$OPENJ9_CPU" = xaarch64 ; then
+    OPENJ9_PLATFORM_CODE=xr64
   else
     as_fn_error $? "Unsupported OpenJ9 cpu ${OPENJ9_CPU}!" "$LINENO" 5
   fi
@@ -15864,7 +15869,7 @@ $as_echo "no (explicitly disabled)" >&6; }
     OPENJ9_ENABLE_DDR=false
   elif test "x$enable_ddr" = x ; then
     case "$OPENJ9_PLATFORM_CODE" in
-      ap64|oa64|wa64|wi32|xa64|xl64|xz64)
+      ap64|oa64|wa64|wi32|xa64|xl64|xr64|xz64)
         { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes (default for $OPENJ9_PLATFORM_CODE)" >&5
 $as_echo "yes (default for $OPENJ9_PLATFORM_CODE)" >&6; }
         OPENJ9_ENABLE_DDR=true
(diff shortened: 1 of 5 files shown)


knn-k:
Ready for review.

jdk/src/solaris/bin/aarch64/jvm.cfg
```diff
@@ -0,0 +1,39 @@
...
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+# ===========================================================================
+# (c) Copyright IBM Corp. 2013, 2020 All Rights Reserved
```
keithc-ca, line 24: The first date here should be 2020 (when this file is added).

knn-k: Changed the copyright year following the comment above.

jdk/make/lib/SoundLibraries.gmk
```diff
@@ -151,6 +151,10 @@ else
     LIBJSOUND_CFLAGS += -DX_ARCH=X_S390X
   endif
 
+  ifeq ($(OPENJDK_TARGET_CPU), aarch64)
+    LIBJSOUND_CFLAGS += -DX_ARCH=X_AARCH64
```
keithc-ca, line 155: Where is `X_AARCH64` defined?
Perhaps `X_AARCH64` (and `X_S390X`) should be added to `SoundDefs.h`?

jdk/make/closed/autoconf/generated-configure.sh
```diff
@@ -15196,6 +15196,9 @@ fi
     powerpc64)
       OPENJ9_CPU=ppc-64
       ;;
+    aarch64)
```
keithc-ca, line 15199: This needs to be regenerated: it's missing the addition of `xr64` in the list of acceptable values of `OPENJ9_PLATFORM_CODE`.

jdk/make/lib/SoundLibraries.gmk
knn-k: Something needs to be passed to `X_ARCH`, otherwise you see the error message "You need to define X_PLATFORM and X_ARCH outside of the source."
The `X_ARCH` value is not used at all, however.

keithc-ca: I see uses in `Platform.c`, e.g. https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/native/com/sun/media/sound/Platform.c#L47.

jdk/make/closed/autoconf/generated-configure.sh
knn-k: Sorry, but I don't understand what you mean.
Where is the list of acceptable values of `OPENJ9_PLATFORM_CODE`?

keithc-ca: That's exactly my point: it's not here (but it should be). The change to line 181 of `custom-hook.m4` is not reflected in this file.

jdk/make/lib/SoundLibraries.gmk
knn-k: SoundDefs.h has no line for IBM copyright.  Do I need to add the line?  What would be the first year in the line in that case?

knn-k: I added the definitions of `X_S390X` and `X_AARCH64` in SoundDefs.h.

jdk/make/closed/autoconf/generated-configure.sh
knn-k: I added `xr64` to line 15872 of generated-configure.sh.

keithc-ca: Normally, you would just re-run `bash common/autoconf/autogen.sh`, but the result your way is fine.

keithc-ca:
Jenkins compile zlinux jdk8

PR #315, opened 2019-06-13
WIP: Add JITServer conditional build flag

a7ehuo:
Setting `—enable-jitserver=yes` will export
`JITSERVER_SUPPORT=1` to the JIT build environment. 
Without `—enable-jitserver` or explicitly setting
`—enable-jitserver=no` will not export `JITSERVER_SUPPORT=1`.

Signed-off-by: Annabelle Huo <Annabelle.Huo@ibm.com>


mstoodle:
Maybe this is a moment we should start thinking harder about using the word 'jitaas' ?

mpirvu:
Personally I am ok with "JITaaS" because we advertised it as such to the community in all our presentations. Even twitter has #JITaaS. A google search does not indicate any hits on other strange things.

mpirvu:
Possible alternative names as suggested by @mstoodle:  jitservice or jitsrvc or jitserv
Yet another possibility: OOPJIT (from out-of-process JIT)

mstoodle:
Let me add a little more reasoning: I think "JITaaS" is too hard to pronounce unambiguously: two A's and one S can too easily to sound like one A and two S's ;) . Call me a prude, I guess...

DanHeidinga:
I second @mstoodle's concerns.  Think about the name like a elementary school student looking to mock it.  `JITaaS` is just too easy to mock.

Let's pick something else that's less mockable and presents a better image to the users

a7ehuo:
As a side note, we should make sure the flags are also picked up by CMake when we revisit this PR after `JITServer` code is merged into the `master` branch.

PR #316, opened 2019-06-18
Allow multiple possible library paths

adamfarley:
Currently, if the cumulative total length of all possible library
paths is greater than the maximum length of a single path (on this
OS), then the path to the library is left blank. This results in
failure to load the library.

This fix is the all-platforms (other than windows) port for the PR
linked below, and enables each separate path to be measured for
length independent of its neighbours.

https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/107

### DIFF
--- a/jdk/src/share/back/export/sys.h
+++ b/jdk/src/share/back/export/sys.h
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
+ * ===========================================================================
+ */
+
 #ifndef JDWP_SYS_H
 #define JDWP_SYS_H
 
@@ -37,7 +43,7 @@
 
 /* Implemented in linker_md.c */
 
-void    dbgsysBuildLibName(char *, int, const char *, const char *);
+void    dbgsysBuildLibName(char *, size_t, const char *, const char *);
 void *  dbgsysLoadLibrary(const char *, char *err_buf, int err_buflen);
 void    dbgsysUnloadLibrary(void *);
 void *  dbgsysFindLibraryEntry(void *, const char *);
--- a/jdk/src/solaris/back/linker_md.c
+++ b/jdk/src/solaris/back/linker_md.c
@@ -24,6 +24,12 @@
  */
 
 /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
+ * ===========================================================================
+ */
+
+/*
  * Adapted from JDK 1.2 linker_md.c v1.37. Note that we #define
  * NATIVE here, whether or not we're running solaris native threads.
  * Outside the VM, it's unclear how we can do the locking that is
@@ -57,6 +63,7 @@
 static void dll_build_name(char* buffer, size_t buflen,
                            const char* paths, const char* fname) {
     char *path, *paths_copy, *next_token;
+    *buffer = '\0';
 
     paths_copy = strdup(paths);
     if (paths_copy == NULL) {
@@ -67,8 +74,10 @@ static void dll_build_name(char* buffer, size_t buflen,
     path = strtok_r(paths_copy, PATH_SEPARATOR, &next_token);
 
     while (path != NULL) {
-        snprintf(buffer, buflen, "%s/lib%s." LIB_SUFFIX, path, fname);
-        if (access(buffer, F_OK) == 0) {
+        size_t result_len = (size_t)snprintf(buffer, buflen, "%s/lib%s." LIB_SUFFIX, path, fname);
+        if (result_len >= buflen) {
+            /* Ignore this path that doesn't fit in the supplied buffer. */
+        } else if (access(buffer, F_OK) == 0) {
             break;
         }
         *buffer = '\0';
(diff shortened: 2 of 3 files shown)


adamfarley:
The bug this PR fixes can be shown by running this command, where $toolong is a fake path which is beyond the maximum path length for that OS, and $libshere is the correct location of the library jdwp tries to find.

```java -Xrunjdwp:transport=dt_socket,address=8888,server=y,onthrow=no.pkg.foo,launch=echo -Dsun.boot.library.path=$toolong:$libshere TestClass```

P.S. TestClass prints a message saying it has run, and does nothing else.

File: jdk/src/solaris/back/linker_md.c [hunk 1]
```diff
@@ -22,6 +22,12 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+
```
keithc-ca: Please remove the trailing whitespace.

jdk/src/solaris/back/linker_md.c
```diff
@@ -101,13 +109,13 @@ dbgsysBuildLibName(char *holder, int holderlen, const char *pname, const char *f
     const int pnamelen = pname ? strlen(pname) : 0;
 
     *holder = '\0';
```
keithc-ca: This assignment is now redundant: please remove it.

jdk/src/solaris/back/linker_md.c
adamfarley: Done.

jdk/src/solaris/back/linker_md.c
adamfarley: Done.

adamfarley:
@keithc-ca - if you're happy with the changes, can you approve and merge this PR, as well as the counterparts?

jdk11 - https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/189
jdk12 - https://github.com/ibmruntimes/openj9-openjdk-jdk12/pull/58
jdk13 - https://github.com/ibmruntimes/openj9-openjdk-jdk13/pull/1
jdk head - https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/121

jdk/src/solaris/back/linker_md.c
```diff
@@ -100,14 +108,12 @@ dbgsysBuildLibName(char *holder, int holderlen, const char *pname, const char *f
 {
     const int pnamelen = pname ? strlen(pname) : 0;
 
-    *holder = '\0';
```
keithc-ca: I had assumed that `dll_build_name` would do the equivalent of this on failure, but as I look at it again, it does not. Could you fix that?

adamfarley: Done. I could have put it in the if, but this way we're covered even if strtok_s fails and returns a null, or if anything else unforeseen happens.

keithc-ca:
Jenkins compile win32 jdk8

keithc-ca:
Jenkins compile win32 jdk8

adamfarley:
Re-running checks, as the size_t-int change has been applied to all versions of the jdk for consistency.

adamfarley:
Jenkins compile win64 jdk8

adamfarley:
Jenkins compile win jdk8

adamfarley:
Jenkins compile xlinux jdk8

jdk/src/solaris/back/linker_md.c
```diff
@@ -54,9 +60,10 @@
 #define LIB_SUFFIX "so"
 #endif
 
-static void dll_build_name(char* buffer, size_t buflen,
+static void dll_build_name(char* buffer, int buflen,
```
keithc-ca: This is inappropriate: `size_t` is the right type to use here.

adamfarley: There was me trying to avoid a similar error in dll_build_name. I presumed it'd "warn" me when it tried compiling the narrowing conversion later on (in the dll_build_name call).

Change in. Running build.

adamfarley:
Jenkins compile xlinux jdk8

jdk/src/share/back/export/sys.h
```diff
@@ -37,7 +37,7 @@
 
 /* Implemented in linker_md.c */
 
-void    dbgsysBuildLibName(char *, int, const char *, const char *);
+void    dbgsysBuildLibName(char *, size_t, const char *, const char *);
```
keithc-ca: This file needs the IBM copyright added.

adamfarley: Done.

adamfarley:
There are only two implementations of dbgsysBuildLibName. Both are changed.

The methods that call it still pass an int, but those files were not changed because they're passing the results of a "sizeof", which always returns a positive int.

keithc-ca:
This looks good. Please squash before I start testing.

adamfarley:
Done.

keithc-ca:
> they're passing the results of a "sizeof", which always returns a positive int.

The value is a positive integer, but not strictly an `int`; see [1] and [2] which note that the result type of the `sizeof` operator is `size_t`.

[1] https://en.wikipedia.org/wiki/Sizeof
[2] https://en.cppreference.com/w/cpp/language/sizeof

keithc-ca:
Jenkins compile xlinux,win,win32 jdk8

keithc-ca:
Please update the related pull requests according to the pattern of this one.

adamfarley:
@keithc-ca - The other pull requests have been updated.

PR #317, opened 2019-06-18
Set the close state of the classpaths

ChengJin01:
The change is to notify the close state
of the classpaths when the corresponding
URLClassloader is closed.

Signed-off-by: Cheng Jin <jincheng@ca.ibm.com>

### DIFF
--- a/jdk/src/share/classes/java/net/URLClassLoader.java
+++ b/jdk/src/share/classes/java/net/URLClassLoader.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2000, 2018 All Rights Reserved
+ * (c) Copyright IBM Corp. 2000, 2019 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


ChengJin01:
Reviewer: @pshipton , @hangshao0
FYI: @DanHeidinga

ChengJin01:
Note: the changes must be merged with https://github.com/eclipse/openj9/pull/6174

pshipton:
I'm not convinced this change should be added, I don't think the feature works properly. If the jar file is used by multiple classloaders, removing the CPM_ZIP_OPEN flag will cause timestamp checks to start occurring every time for the remaining classloaders using the file.

ChengJin01:
> I'm not convinced this change should be added, I don't think the feature works properly. If the jar file is used by multiple classloaders, removing the CPM_ZIP_OPEN flag will cause timestamp checks to start occurring every time for the remaining classloaders using the file.

It makes more sense to me at this point. Then all changes related to the close state should be removed.

ChengJin01:
Close the PR as it is incorrect to remove CPM_ZIP_OPEN as explained above.

PR #321, opened 2019-07-08
Fix test image rule to use correct path

fengxue-IS:
Signed-off-by: Jack Lu <Jack.S.Lu@ibm.com>


DanHeidinga:
port from jdk.next: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/125

PR #325, opened 2019-08-22
WIP: Fix javadoc for openj9 classes

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

PR #342, opened 2019-10-30
Add targets to create openj9 javadoc

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

### DIFF
--- a/closed/make/Javadoc.gmk
+++ b/closed/make/Javadoc.gmk
@@ -1,0 +1,485 @@
+# Copyright (c) 1997, 2015, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  Oracle designates this
+# particular file as subject to the "Classpath" exception as provided
+# by Oracle in the LICENSE file that accompanied this code.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+#
+
+# ===========================================================================
+# (c) Copyright IBM Corp. 2019, 2020 All Rights Reserved
+# ===========================================================================
+
+#############################################################################
+# This file is included by (and is based on) make/Javadoc.gmk
+# OpenJ9 adds a number of packages to the jdk and also overrides certain
+# core api and com.sun.management classes.
+# The targets in this file generate the javadoc for the additional classes
+# and modify the page footers in the overridden files to carry the correct
+# copyrights.
+# Javadoc for the OpenJ9 extensions packages is created under docs/platform
+#############################################################################
+
+#############################################################################
+# Set the link prefix for the JavaSE core and com.sun.management packages
+# to reference the pages created in this build.
+JAVASE_BASE_URL := $(JDKJRE2COREAPI)
+
+#############################################################################
+# Source files for openj9 extensions javadoc
+OPENJ9_DOCS_SRCDIR := $(JDK_OUTPUTDIR)/j9jcl_sources/jdk/src/share/classes
+
+#############################################################################
+# Update macros which hold a list of the source directories.
+# ALL_SOURCE_DIRS is used in the PackageFilter function
+ALL_SOURCE_DIRS := $(OPENJ9_DOCS_SRCDIR) $(ALL_SOURCE_DIRS)
+# ALL_EXISTING_SOURCE_DIRS is used in the $(DIRECTORY_CACHE) target
+ALL_EXISTING_SOURCE_DIRS := $(wildcard $(ALL_SOURCE_DIRS))
+
+#############################################################################
+# html page footer for the OpenJ9 extensions and overrides pages
+OPENJ9_BASE_URL := https://www.eclipse.org/openj9/
+OPENJ9_BUG_SUBMIT_URL := https://github.com/eclipse/openj9/issues
+OPENJ9_FULL_COMPANY_NAME := IBM Corp
+OPENJ9_JAVADOC_BOTTOM = \
+    <a href="$(OPENJ9_BASE_URL)">Link</a> to Eclipse OpenJ9 documentation.<br> \
+    To raise a bug report or suggest an improvement create an <a href="$(OPENJ9_BUG_SUBMIT_URL)">Eclipse Openj9 issue.</a><br> \
+    Copyright &copy; $(FIRST_COPYRIGHT_YEAR), $(COPYRIGHT_YEAR), $(OPENJ9_FULL_COMPANY_NAME). All rights reserved.
+
+#############################################################################
+#
+# openj9-sharedclassesdocs
+#
+
+OPENJ9_SHAREDCLASSES_PKGS := com.ibm.oti.shared
+
+OPENJ9_SHAREDCLASSES_FIRST_COPYRIGHT_YEAR := 1998
+
+OPENJ9_SHAREDCLASSES_DOCDIR := $(PLATFORM_DOCSDIR)/sharedclasses
+OPENJ9_SHAREDCLASSES_DOCTITLE := OpenJ9 Shared Classes
+OPENJ9_SHAREDCLASSES_WINDOWTITLE := OpenJ9 Shared Classes
+OPENJ9_SHAREDCLASSES_HEADER := <strong>OpenJ9 Shared Classes</strong>
+
+# The index.html, options, and packages files
+OPENJ9_SHAREDCLASSES_INDEX_FILE := $(OPENJ9_SHAREDCLASSES_DOCDIR)/index.html
+OPENJ9_SHAREDCLASSES_OPTIONS_FILE := $(DOCSTMPDIR)/openj9.sharedclasses.options
+OPENJ9_SHAREDCLASSES_PACKAGES_FILE := $(DOCSTMPDIR)/openj9.sharedclasses.packages
+
+openj9-sharedclassesdocs: $(OPENJ9_SHAREDCLASSES_INDEX_FILE)
+
+# Set relative location to platform document root
+$(OPENJ9_SHAREDCLASSES_INDEX_FILE): GET2DOCSDIR := $(PLATFORM_DOCSDIR)/..
+
+# Run javadoc if the index file is out of date or missing
+$(OPENJ9_SHAREDCLASSES_INDEX_FILE): $(OPENJ9_SHAREDCLASSES_OPTIONS_FILE) $(OPENJ9_SHAREDCLASSES_PACKAGES_FILE)
+	$(prep-javadoc)
+	$(call JavadocSummary,$(OPENJ9_SHAREDCLASSES_OPTIONS_FILE),$(OPENJ9_SHAREDCLASSES_PACKAGES_FILE))
+	$(JAVADOC_CMD) -d $(@D) \
+	    @$(OPENJ9_SHAREDCLASSES_OPTIONS_FILE) @$(OPENJ9_SHAREDCLASSES_PACKAGES_FILE)
+
+$(OPENJ9_SHAREDCLASSES_OPTIONS_FILE): FIRST_COPYRIGHT_YEAR := $(OPENJ9_SHAREDCLASSES_FIRST_COPYRIGHT_YEAR)
+$(OPENJ9_SHAREDCLASSES_OPTIONS_FILE):
+	$(prep-target)
+	@($(call OptionOnly,$(COMMON_JAVADOCFLAGS)) ; \
+	  $(call OptionPair,-sourcepath,$(OPENJ9_DOCS_SRCDIR)) ; \
+	  $(call OptionPair,-encoding,ascii) ; \
+	  $(call OptionPair,-doctitle,$(OPENJ9_SHAREDCLASSES_DOCTITLE)) ; \
+	  $(call OptionPair,-windowtitle,$(OPENJ9_SHAREDCLASSES_WINDOWTITLE) $(DRAFT_WINTITLE)) ; \
+	  $(call OptionPair,-header,$(OPENJ9_SHAREDCLASSES_HEADER)$(DRAFT_HEADER)) ; \
+	  $(call OptionPair,-bottom,$(OPENJ9_JAVADOC_BOTTOM)$(DRAFT_BOTTOM)) ; \
+	  $(call OptionTrip,-linkoffline,$(JAVASE_BASE_URL),$(COREAPI_DOCSDIR)/) ; \
+	) > $@
+
+# Create a file with the package names in it
+$(OPENJ9_SHAREDCLASSES_PACKAGES_FILE): $(DIRECTORY_CACHE) $(call PackageDependencies,$(OPENJ9_SHAREDCLASSES_PKGS))
+	$(prep-target)
+	$(call PackageFilter,$(OPENJ9_SHAREDCLASSES_PKGS))
+
+#############################################################################
+#
+# openj9-gpudocs
+#
+
+OPENJ9_GPU_PKGS := com.ibm.gpu
+
+OPENJ9_GPU_FIRST_COPYRIGHT_YEAR := 2014
+
+OPENJ9_GPU_DOCDIR := $(PLATFORM_DOCSDIR)/gpu
+OPENJ9_GPU2COREAPI := $(JDKJRE2COREAPI)
+OPENJ9_GPU_DOCTITLE := OpenJ9 GPU Classes
+OPENJ9_GPU_WINDOWTITLE := OpenJ9 GPU Classes
+OPENJ9_GPU_HEADER := <strong>OpenJ9 GPU Classes</strong>
+
+# The index.html, options, and packages files
+OPENJ9_GPU_INDEX_FILE := $(OPENJ9_GPU_DOCDIR)/index.html
+OPENJ9_GPU_OPTIONS_FILE := $(DOCSTMPDIR)/openj9.gpu.options
+OPENJ9_GPU_PACKAGES_FILE := $(DOCSTMPDIR)/openj9.gpu.packages
+
+openj9-gpudocs: $(OPENJ9_GPU_INDEX_FILE)
+
+# Set relative location to platform document root
+$(OPENJ9_GPU_INDEX_FILE): GET2DOCSDIR := $(PLATFORM_DOCSDIR)/..
+
+# Run javadoc if the index file is out of date or missing
+$(OPENJ9_GPU_INDEX_FILE): $(OPENJ9_GPU_OPTIONS_FILE) $(OPENJ9_GPU_PACKAGES_FILE)
+	$(prep-javadoc)
+	$(call JavadocSummary,$(OPENJ9_GPU_OPTIONS_FILE),$(OPENJ9_GPU_PACKAGES_FILE))
+	$(JAVADOC_CMD) -d $(@D) \
+	    @$(OPENJ9_GPU_OPTIONS_FILE) @$(OPENJ9_GPU_PACKAGES_FILE)
+
+$(OPENJ9_GPU_OPTIONS_FILE): FIRST_COPYRIGHT_YEAR := $(OPENJ9_GPU_FIRST_COPYRIGHT_YEAR)
+$(OPENJ9_GPU_OPTIONS_FILE):
+	$(prep-target)
+	@($(call OptionOnly,$(COMMON_JAVADOCFLAGS)) ; \
+	  $(call OptionPair,-sourcepath,$(OPENJ9_DOCS_SRCDIR)) ; \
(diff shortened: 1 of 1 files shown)


lumpfish:
This is a follow on from https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/325.  This PR uses does not require an additional plugin point for a new closed/make/Javadoc-post file - instead it creates the same javadoc using the existing plugin point for closed/make/Javadoc.gmk.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+	echo OPENJ9_SHAREDCLASSES2COREAPI is $(OPENJ9_SHAREDCLASSES2COREAPI)
+	@($(call OptionOnly,$(COMMON_JAVADOCFLAGS)) ; \
+	  $(call OptionPair,-sourcepath,$(OPENJ9_DOCS_SRCDIR)) ; \
+	  $(call OptionPair,-encoding,ascii) ; \
+	  $(call OptionPair,-doctitle,$(OPENJ9_SHAREDCLASSES_DOCTITLE)) ; \
+	  $(call OptionPair,-windowtitle,$(OPENJ9_SHAREDCLASSES_WINDOWTITLE) $(DRAFT_WINTITLE)); \
+	  $(call OptionPair,-header,$(OPENJ9_SHAREDCLASSES_HEADER)$(DRAFT_HEADER)) ; \
+	  $(call OptionPair,-bottom,$(OPENJ9_JAVADOC_BOTTOM)$(DRAFT_BOTTOM)) ; \
+	  $(call OptionTrip,-linkoffline,$(JAVASE_BASE_URL),$(COREAPI_DOCSDIR)/); \
+	) >> $@
```
keithc-ca: Why does this append to the target file rather than just write to it (`>>` instead of `>`)?

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+OPENJ9_SHAREDCLASSES_OPTIONS_FILE := $(DOCSTMPDIR)/openj9.sharedclasses.options
+OPENJ9_SHAREDCLASSES_PACKAGES_FILE := $(DOCSTMPDIR)/openj9.sharedclasses.packages
+
+openj9-sharedclassesdocs: $(OPENJ9_SHAREDCLASSES_INDEX_FILE)
+
+# Set relative location to platform document root
+$(OPENJ9_SHAREDCLASSES_INDEX_FILE): GET2DOCSDIR := $(PLATFORM_DOCSDIR)/..
+
+# Run javadoc if the index file is out of date or missing
+$(OPENJ9_SHAREDCLASSES_INDEX_FILE): $(OPENJ9_SHAREDCLASSES_OPTIONS_FILE) $(OPENJ9_SHAREDCLASSES_PACKAGES_FILE) coredocs
```
keithc-ca: Because `coredocs` is a phony target, this rule will alway run; likewise for other packages.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+	$(prep-target)
+	@($(call OptionOnly,$(COMMON_JAVADOCFLAGS)) ; \
+	  $(call OptionPair,-sourcepath,$(OPENJ9_MGMT_SRCPATH)) ; \
+	  $(call OptionPair,-encoding,ascii) ; \
+	  $(call OptionPair,-doctitle,$(MGMT_DOCTITLE)) ; \
+	  $(call OptionPair,-windowtitle,$(MGMT_WINDOWTITLE) $(DRAFT_WINTITLE)); \
+	  $(call OptionPair,-header,$(MGMT_HEADER)$(DRAFT_HEADER)) ; \
+	  $(call OptionPair,-bottom,$(MGMT_BOTTOM)$(DRAFT_BOTTOM)) ; \
+	  $(call OptionTrip,-linkoffline,$(JAVASE_BASE_URL),$(COREAPI_DOCSDIR)/); \
+	  $(call OptionTrip,-linkoffline,$(MGMT_BASE_URL),$(MGMT_DOCDIR)/); \
```
keithc-ca: Inconsistent spacing: missing space before `;` to match lines above.

closed/make/Javadoc.gmk
lumpfish: Just because that's what the similar targets in make/Javadoc.gmk do.  Doesn't alter the output to use '>', so I've changed it for the targets in this file.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+    <a href="$(OPENJ9_BASE_URL)">Link</a> to Eclipse OpenJ9 documentation.<br> \
+    To raise a bug report or suggest an improvement create an <a href="$(OPENJ9_BUG_SUBMIT_URL)">Eclipse Openj9 issue.</a><br> \
+    Copyright &copy; $(FIRST_COPYRIGHT_YEAR), $(COPYRIGHT_YEAR), $(OPENJ9_FULL_COMPANY_NAME). All rights reserved.
+
+#############################################################
+#
+# openj9-sharedclassesdocs
+#
+
+OPENJ9_SHAREDCLASSES_PKGDIR := $(OPENJ9_TOPDIR)/jcl/src/openj9.sharedclasses/share/classes
```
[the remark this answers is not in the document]
lumpfish: Updated to use j9jcl_sources.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+
+#############################################################
+#
+# openj9-gpudocs
+#
+
+OPENJ9_GPU_PKGDIR := $(OPENJ9_TOPDIR)/jcl/src/openj9.gpu/share/classes/
+OPENJ9_GPU_PKGS := com.ibm.gpu
+
+OPENJ9_GPU_FIRST_COPYRIGHT_YEAR=2014
```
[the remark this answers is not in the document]
lumpfish: Fixed.

closed/make/Javadoc.gmk
lumpfish: Removed it as a dependency for the open9-xxx targets, but since openj9-mgmtdocs depends on the openjdk mgmtdocs which does have a dependency on coredocs, it will run anyway.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+$(OPENJ9_GPU_INDEX_FILE): GET2DOCSDIR := $(PLATFORM_DOCSDIR)/..
+
+# Run javadoc if the index file is out of date or missing
+$(OPENJ9_GPU_INDEX_FILE): $(OPENJ9_GPU_OPTIONS_FILE) $(OPENJ9_GPU_PACKAGES_FILE) coredocs
+	$(prep-javadoc)
+	$(call JavadocSummary,$(OPENJ9_GPU_OPTIONS_FILE),$(OPENJ9_GPU_PACKAGES_FILE))
+	$(JAVADOC_CMD) -d $(@D) \
+	    @$(OPENJ9_GPU_OPTIONS_FILE) @$(OPENJ9_GPU_PACKAGES_FILE)
+
+$(OPENJ9_GPU_OPTIONS_FILE): FIRST_COPYRIGHT_YEAR=$(OPENJ9_GPU_FIRST_COPYRIGHT_YEAR)
```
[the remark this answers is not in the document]
lumpfish: In addition to resolving the review comments, the PR has been updated to update the page footer for the core api javadoc pages for the classes which are overridden by openj9, by mapping the source files in j9jcl_sources to their corresponding javadoc html files.

closed/make/Javadoc.gmk
lumpfish: Fixed

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,475 @@
...
+	# Replace footer for com.ibm packages and classes
+	$(CD) $(MGMT_DOCDIR); find . -type f -name "*.html" | grep "/com/ibm" | xargs -I {} $(OPENJ9_MGMT_SUBST_COMMAND) {}
+	# Replace footer for the com.sun.management classes reimplemented in OpenJ9
+	$(CD) $(MGMT_DOCDIR); find . -type f -name "GarbageCollectionNotificationInfo.html" | xargs -I {} $(OPENJ9_MGMT_SUBST_COMMAND) {}
+	$(CD) $(MGMT_DOCDIR); find . -type f -name "GcInfo.html" | xargs -I {} $(OPENJ9_MGMT_SUBST_COMMAND) {}
+
+ALL_OTHER_TARGETS += openj9-docs
+openj9-docs: openj9-sharedclassesdocs openj9-gpudocs openj9-cudadocs openj9-dataaccessdocs openj9-jvmdocs openj9-mgmtdocs
+
+.PHONY: openj9-sharedclassesdocs openj9-gpudocs openj9-cudadocs openj9-dataaccessdocs openj9-jvmdocs openj9-mgmtdocs openj9-mgmtdocs_prepare openj9-mgmtdocs-restore
```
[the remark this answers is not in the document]
lumpfish: The targets in ALL_OTHER_TARGETS are declared as PHONY in make/Javadoc.gmk, so those are not required to be declared phony in this file.  Added a comment to that effect.

lumpfish:
The page footer replace code has been updated to be similar to that for jdk11 in https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/212

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,490 @@
...
+        $(CP) "$(JRE_API_DOCSDIR)/management/extension/$$path" "$(OPENJ9_JAVADOC_TEMP)" ; \
+        $(OPENJ9_MGMT_SUBST_COMMAND) < "$(OPENJ9_JAVADOC_TEMP)" > "$(JRE_API_DOCSDIR)/management/extension/$$path" ; \
+        $(RM) "$(OPENJ9_JAVADOC_TEMP)" ; \
+      fi ; \
+    done \
+    #
+
+$(OPENJ9_JAVADOC_MARKER) : $(COREAPI_DOCSDIR)/index.html openj9-mgmtdocs
+	$(ECHO) '$(OPENJ9_JAVADOC_SCRIPT_BODY)' >$(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(SUPPORT_OUTPUTDIR)/j9jcl_sources ; \
```
keithc-ca: The directory should be `$(JDK_OUTPUTDIR)/j9jcl_sources`.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,490 @@
...
+		$(FIND) . -type f -name '*.java' \
+			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(TOPDIR)/closed/src ; \
```
keithc-ca: The directory should be `$(TOPDIR)/closed/adds/jdk/src`.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,490 @@
...
+      fi ; \
+      if [ -f "$(JRE_API_DOCSDIR)/management/extension/$$path" ] ; then \
+        $(CP) "$(JRE_API_DOCSDIR)/management/extension/$$path" "$(OPENJ9_JAVADOC_TEMP)" ; \
+        $(OPENJ9_MGMT_SUBST_COMMAND) < "$(OPENJ9_JAVADOC_TEMP)" > "$(JRE_API_DOCSDIR)/management/extension/$$path" ; \
+        $(RM) "$(OPENJ9_JAVADOC_TEMP)" ; \
+      fi ; \
+    done \
+    #
+
+$(OPENJ9_JAVADOC_MARKER) : $(COREAPI_DOCSDIR)/index.html openj9-mgmtdocs
```
keithc-ca: `openj9-mgmtdocs` appears to be a phony target (which should explicitly be declared). Using a phony target as a prerequisite will force execution of this rule unconditionally.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,490 @@
...
+	$(ECHO) '$(OPENJ9_JAVADOC_SCRIPT_BODY)' >$(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(SUPPORT_OUTPUTDIR)/j9jcl_sources ; \
+		$(FIND) . -type f -name '*.java' \
+			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(TOPDIR)/closed/src ; \
+		$(FIND) . -type f -name '*.java' \
+			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(COREAPI_DOCSDIR) ; \
```
keithc-ca: Do any of `JDK_API_DOCSDIR`, `JRE_API_DOCSDIR` or `PLATFORM_DOCSDIR` need to be processed?

closed/make/Javadoc.gmk
lumpfish: Apologies, I had neglected to check in the final version of the code.  Those directories are now fixed.

closed/make/Javadoc.gmk
lumpfish: Yes, sorry, fixed now.

closed/make/Javadoc.gmk
lumpfish: openj9-mgmtdocs is declared as PHONY in make/Javadoc.gmk as a result of it being included in ALL_OTHER_TARGETS.
It was done this way simply because that is the model used by the openjdk build for the various doc areas - a phony target with a dependency on the index.html file for that area.

closed/make/Javadoc.gmk
lumpfish: The code under review was incorrect and should now be fixed.
But to answer the question:
JDK_API_DOCSDIR - contains javadoc for attach, javac, javadoc, jconsole, jpda, nashorn.  I do not believe these APIs are modified by openj9
JRE_API_DOCSDIR - contains javadoc for accessibility, management, net, nio, plugin, security.  We need to process JRE_API_DOCSDIR/management/extension because openj9 provides its own management classes
PLATFORM_DOCSDIR - contains javadoc for the openj9 specific packages, but we don't need to post process these files because the footer is set correctly on the javadoc commands (the issue with some files needing openjdk footers and some needing openj9 footers doesn't arise).

keithc-ca: I believe there are openj9 contributions to all of those groups:
* jcmd, jps should appear in JDK_API_DOCSDIR
* JRE_API_DOCSDIR should include java.base classes - see the next point
* if any java.base classes appear in JDK_API_DOCSDIR, they may need fixing (e.g. java.lang.String)

closed/make/Javadoc.gmk
keithc-ca: I'm fine if that target is declared PHONY elsewhere, but the unconditional execution of this rule is a problem that needs to be fixed.

closed/make/Javadoc.gmk
lumpfish: 1. javadoc is not produced for jcmd and jps by the openjdk build - that is why we no not see the javadoc for the openj9 jdk.  
2. The base class modifications are included in the coredocs javadoc (docs/api).  Only the openj9.xx packages are included under PLATFORM_DOCSDIR.

closed/make/Javadoc.gmk
lumpfish: My assumption is that architected way for additional javadoc targets to be added to the build is to add them to the list of targets in the ALL_OTHER_TARGETS macro, which are dependencies of the 'otherdocs' target which is a dependency of the 'docs' target.  The ALL_OTHER_TARGETS  targets are all declared as phony in the including make/Javadoc.gmk, so they are all phony dependencies.
Or have I misunderstood your point?

keithc-ca: It's fine for PHONY targets to depend on real files (as long as the script for those phony targets does nothing substantial), but not the other way around.

closed/make/Javadoc.gmk
keithc-ca: I didn't do a thorough investigation, but noticed no OpenJ9 footers anywhere other than in JDK_API_DOCSDIR. That's fine if there shouldn't be anything there, but it was surprising, and I would expect to se jcmd and jps, for example, documented somewhere.

I guess the summary, from my point of view, is that I don't see the structure in the makefiles to deal with future changes to Java code in OpenJ9 that should land in those extra folders. Please help me see it or add it.

lumpfish: The openj9 overrides to the core packages replace the openjdk ones under docs/api - so for instance docs/api/java/lang/String.html contains the javadoc and the footer from the openj9 String class override.
The sources files are identified by the find command:
```
	$(CD) $(JDK_OUTPUTDIR)/j9jcl_sources/jdk/src ; \
		$(FIND) . -type f -name '*.java' \
			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
```
and the corresponding html file footers replaced by the
```
if [ -f "$(COREAPI_DOCSDIR)/$$path" ] ; then \
        $(CP) "$(COREAPI_DOCSDIR)/$$path" "$(OPENJ9_JAVADOC_TEMP)" ; \
        $(OPENJ9_COREAPI_SUBST_COMMAND) < "$(OPENJ9_JAVADOC_TEMP)" > "$(COREAPI_DOCSDIR)/$$path" ; \
        $(RM) "$(OPENJ9_JAVADOC_TEMP)" ; \
      fi ;
```
statement in the temporary script file.

closed/make/Javadoc.gmk
lumpfish: OK.  So my issue is how to use the ALL_OTHER_TARGETS as a means of adding the extra javadoc targets to the build, but to be sure of only executing the target to replace the footers when the core api javadoc and the management/extensions javadoc generation has completed.
The code currently does that by depending on openj9_mgmtdocs, which depends on mgmtdocs (in the openjdk make/Javadoc.gmk file) which depends on coredocs (in the openjdk make/Javadoc.gmk file).

lumpfish: I've replaced the phony target dependencies with the file targets which those phony targets depended on.  @keithc-ca - please can you take another look.

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,486 @@
...
+	) > $@
+
+# Create a file with the package names in it
+$(OPENJ9_JVM_PACKAGES_FILE): $(DIRECTORY_CACHE) $(call PackageDependencies,$(OPENJ9_JVM_PKGS))
+	$(prep-target)
+	$(call PackageFilter,$(OPENJ9_JVM_PKGS))
+
+#############################################################################
+#
+# openj9 mgmtdocs
```
keithc-ca: Mostly a curiosity: why did you replace the hyphen with a space here (all similar comments still use a hyphen)?

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,486 @@
...
+
+#############################################################################
+# Location of the openjdk com.sun.management source files
+SUN_MGMT_SRCDIR := $(JDK_SHARE_CLASSES)/com/sun/management
+
+#############################################################################
+# Somewhere to copy the openjdk com.sun.management interfaces referenced by
+# the com.ibm.lang.management to.
+TEMP_SUN_MGMT_PKGDIR := $(TEMPDIR)/mgmt
+TEMP_SUN_MGMT_SRCDIR := $(TEMP_SUN_MGMT_PKGDIR)/com/sun/management
```
keithc-ca: How does copying files to this directory help? I can't find a way that `javadoc` is informed of the directory.

closed/make/Javadoc.gmk
lumpfish: In the previous comments (with a hyphen) the comment is also the name of the target which is run.
As a result of removing the phony dependencies in openj9 jdk management docs targets this is not now the case in this section (there is no longer an openj9-mgmtdocs target).

closed/make/Javadoc.gmk
lumpfish: Copying the files to the temporary directory provides the means to exclude the jdk.management files which are not supported by an openj9 jdk.

The directory is added to the openj9 classes directory in the statement
```
OPENJ9_MGMT_SRCPATH := $(OPENJ9_DOCS_SRCDIR)$(PATH_SEP)$(TEMP_SUN_MGMT_PKGDIR)
```
and then passed to javadoc as the -sourcepath argument in the options file:
```
$(OPENJ9_MGMT_OPTIONS_FILE):
	.....
	  $(call OptionPair,-sourcepath,$(OPENJ9_MGMT_SRCPATH)) ; \
	.....
```

closed/make/Javadoc.gmk
```diff
@@ -0,0 +1,486 @@
...
+	  $(call OptionPair,-windowtitle,$(MGMT_WINDOWTITLE) $(DRAFT_WINTITLE)) ; \
+	  $(call OptionPair,-header,$(MGMT_HEADER)$(DRAFT_HEADER)) ; \
+	  $(call OptionPair,-bottom,$(MGMT_BOTTOM)$(DRAFT_BOTTOM)) ; \
+	  $(call OptionTrip,-linkoffline,$(JAVASE_BASE_URL),$(COREAPI_DOCSDIR)/) ; \
+	  $(call OptionTrip,-linkoffline,$(MGMT_BASE_URL),$(MGMT_DOCDIR)/) ; \
+	) > $@
+
+# Create a file with the package names in it
+$(OPENJ9_MGMT_PACKAGES_FILE): $(DIRECTORY_CACHE) $(call PackageDependencies,$(ALL_MGMT_PKGS))
+	echo Running $@
```
keithc-ca: This should use `$(ECHO)` instead of `echo`.

lumpfish: Actually the
```
echo Running $@
```
was a statement left in from debugging.  I'll remove it.

keithc-ca:
Can you please squash now?

lumpfish:
Squashed.

PR #348, opened 2019-11-15
Set default load address for launchers on s390x

M-Davies:
Sets the load address of the launchers to 0x60000 on zlinux 64 systems. This enables compressed references with shift 0 for big heap sizes between 2040m and 4000m.

**Before:**
```
80000000-80001000 r-xp 00000000 fd:01 12062765                           /root/jdk8u232-b09/bin/java
80001000-80002000 r--p 00000000 fd:01 12062765                           /root/jdk8u232-b09/bin/java
80002000-80003000 rw-p 00001000 fd:01 12062765                           /root/jdk8u232-b09/bin/java
```
Screenshot 2019-11-15 at 14 46 34

**After:**
```
00060000-00061000 r-xp 00000000 fd:01 15742334                           /root/SharedDocker/openj9-openjdk-jdk8/build/linux-s390x-normal-server-release/images/j2sdk-image/bin/java
00061000-00062000 r--p 00000000 fd:01 15742334                           /root/SharedDocker/openj9-openjdk-jdk8/build/linux-s390x-normal-server-release/images/j2sdk-image/bin/java
00062000-00063000 rw-p 00001000 fd:01 15742334                           /root/SharedDocker/openj9-openjdk-jdk8/build/linux-s390x-normal-server-release/images/j2sdk-image/bin/java
```
Screenshot 2019-11-15 at 14 48 41

Issue: https://github.com/eclipse/openj9/issues/7115

Signed-off-by: Morgan Davies morgan.davies@ibm.com

### DIFF
--- a/jdk/make/CompileLaunchers.gmk
+++ b/jdk/make/CompileLaunchers.gmk
@@ -93,6 +93,16 @@ define SetupLauncher
   endif
 
   $1_LDFLAGS := $3
+  
+  ifeq ($(OPENJDK_TARGET_OS), linux)
+  # Set the image base address for zLinux 64 to 0x60000 for launchers,
+  # allows compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
+  # / RTC PR 100052
+    ifeq ($(OPENJDK_TARGET_CPU), s390x)
+      $1_LDFLAGS += -Wl,-Ttext-segment=0x60000
+    endif
+  endif
+  
   $1_LDFLAGS_SUFFIX :=
   ifeq ($(OPENJDK_TARGET_OS), macosx)
     $1_PLIST_FILE := Info-cmdline.plist


jdk/make/CompileLaunchers.gmk
```diff
@@ -93,6 +93,18 @@ define SetupLauncher
...
   $1_LDFLAGS := $3
+  
+  ifeq ($(OPENJDK_TARGET_OS), linux)
+  # Set the image base address for zLinux 64 to 0x60000 for launchers,
+  # to make compressedRefsShift to allow compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
+  # / RTC PR 100052
+    ifeq ($(OPENJDK_TARGET_CPU), s390x)
+      ifeq ($(OPENJDK_TARGET_CPU_BITS), 64)
+        $1_LDFLAGS += -Wl,-Ttext-segment=0x60000
+      endif
```
pshipton: I think you have the 32-bit options for this as well, might as well add it, in case it's needed in the future.

jdk/make/CompileLaunchers.gmk
```diff
@@ -93,6 +93,18 @@ define SetupLauncher
   endif
 
   $1_LDFLAGS := $3
+  
+  ifeq ($(OPENJDK_TARGET_OS), linux)
+  # Set the image base address for zLinux 64 to 0x60000 for launchers,
+  # to make compressedRefsShift to allow compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
```
pshipton: seems to be extra words here which can be removed "to make compressedRefsShift"

jdk/make/CompileLaunchers.gmk
M-Davies: In that case, would it be worth just removing the check for 64 bits altogether?

pshipton:
This doesn't "fix" eclipse/openj9#7115 (which is a github keyword) because it needs to be fixed in other Java versions as well. I've updated the description.

Once the comments are addressed, pls squash so there is a single commit.

jdk/make/CompileLaunchers.gmk
pshipton: I thought the 32-bit and 64-bit solutions used different options.

pshipton: Maybe I've confused this with AIX

M-Davies: Oh are you referencing this comment? https://github.com/eclipse/openj9/issues/7115#issuecomment-544535974

If so, I found that it is a AIX only option. This fix should work for 32 bit too but I don't have access to an AIX box to test this theory

pshipton:
Pls update the commit comment. This fix doesn't "enables compressed references" it "enables compressed references **with shift 0**" for bigger heap sizes (from 2040m to about 4000m)

pshipton:
jenkins test sanity zlinux jdk8

PR #355, opened 2020-01-10
CMake: Respect '--enable-cuda' configure option

keithc-ca:
With cmake, `J9VM_OPT_CUDA` is off by default and `OMR_OPT_CUDA` is derived from `J9VM_OPT_CUDA`. We must set `J9VM_OPT_CUDA` on the cmake command line to enable CUDA.


keithc-ca:
Jenkins test sanity xlinux,xlinuxcm jdk8

PR #361, opened 2020-01-17
Stop staging jcl and sourcetools folders

keithc-ca:
* staged jcl folder was not used
* build tools directly from unstaged sourcetools folder
* add missing PHONY declaration for clean-openj9-thirdparty-binaries

Enabled by eclipse/openj9#8246: improves build performance.

PR #362, opened 2020-01-19
CMake: Fix race creating $(OUTPUT_ROOT)/vm

keithc-ca:
There was a race in `OpenJ9.gmk` that apparently since #361 is now more likely to go wrong. That file has these dependencies:
```
run-preprocessors-j9 : $(OUTPUT_ROOT)/vm/cmake.stamp
run-preprocessors-j9 : generate-j9-version-headers
```
A side-effect of `generate-j9-version-headers` is the creation of the directory needed by `$(OUTPUT_ROOT)/vm/cmake.stamp`. This removes the dependency on that ordering.

Also make more like equivalent code in jdk11.


keithc-ca:
Jenkins compile xlinux,xlinuxcm jdk8

PR #366, opened 2020-01-28
Improved AIX performance for multi-thread crypto

shanchao95:
* Add malloc option if AIX

Issue https://github.com/eclipse/openj9/issues/7908

Signed-off-by: Chao Shan <chao.shan@ibm.com>

### DIFF
--- a/jdk/src/solaris/bin/java_md_solinux.c
+++ b/jdk/src/solaris/bin/java_md_solinux.c
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved
+ * ===========================================================================
+ */
+
 #include "java.h"
 #include "jvm_md.h"
 #include <dirent.h>
@@ -380,6 +386,13 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                                            asked for? Current model is
                                            fine unless another model
                                            is asked for */
+#ifdef AIX
+    const char *mallocOptionsName = "MALLOCOPTIONS";
+    const char *mallocOptionsValue = "multiheap,buckets";
+    if (setenv(mallocOptionsName, mallocOptionsValue, 0) != 0) {
+        fprintf(stderr, "setenv('MALLOCOPTIONS=multiheap,buckets') failed: performance may be affected\n");
+    }
+#endif
 #ifdef SETENV_REQUIRED
       jboolean mustsetenv = JNI_FALSE;
       char *runpath     = NULL; /* existing effective LD_LIBRARY_PATH setting */


shanchao95:
eclipse/openj9#7908

jdk/src/solaris/bin/java_md_solinux.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020, Oracle and/or its affiliates. All rights reserved.
```
pshipton: Don't modify the Oracle copyright, pls add an IBM copyright after the Oracle copyright and license.
```
/*
 * ===========================================================================
 * (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved
 * ===========================================================================
 */
```

jdk/src/solaris/bin/java_md_solinux.c
```diff
@@ -380,6 +380,11 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                                            asked for? Current model is
                                            fine unless another model
                                            is asked for */
+#ifdef AIX
+    static char *mallocOptins = "MALLOCOPTIONS=multiheap,buckets";
+    if (putenv(mallocOptins) != 0)
```
pshipton: Pls use { }

pshipton: This overwrites whatever is set in the environment so a user cannot have custom settings, I don't think it should be done this way.

pshipton:
@shanchao95 when you push a change and need a re-review, pls make a comment so the reviewer knows there is something new to look at.

pshipton:
jenkins test sanity,extended,sanity.system,extended.system aix jdk8

jdk/src/solaris/bin/java_md_solinux.c
```diff
@@ -380,6 +386,13 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                                            asked for? Current model is
                                            fine unless another model
                                            is asked for */
+#ifdef AIX
+    static const char *mallocOptinsName = "MALLOCOPTIONS";
+    static const char *mallocOptinsValue = "multiheap,buckets";
+    if (setenv(mallocOptinsName, mallocOptinsValue, 0) != 0) {
+      printf("mallocOptins failed\n");
+    }
```
keithc-ca: The coding style in this repo uses 4 spaces per indentation: please follow suit.

More importantly, this will over-write a user's environment variable: I think this should only set `MALLOCOPTIONS` if it is not already set. I don't think `printf` is the proper response to a `setenv` failure.

Each variable is missing an 'o' in the word 'options'; the should also be declared `static const char []` - we don't need the pointers too.

pshipton: @keithc-ca I don't think `setenv(mallocOptinsName, mallocOptinsValue, 0)`, passing 0 for the overwrite parameter, does overwrite an existing variable.

@shanchao95 note we don't need static vars, as setenv() makes copies of the strings.

shanchao95:
@pshipton @keithc-ca Thanks. The code is updated. 

- use 4 spaces per indentation
- remove static variable, keep the variable type as `const char*` because this is defined argument in function setenv().
- May i get a suggestion about response if `setenv()` fails

jdk/src/solaris/bin/java_md_solinux.c
keithc-ca: Perhaps I was thinking of `putenv` which always overwrites existing values.

I'm fine with inlining the string constants, but if we keep the variables, they should be
```
static const char mallocOptionsName[] = "MALLOCOPTIONS";
static const char mallocOptionsValue[] = "multiheap,buckets";
```

jdk/src/solaris/bin/java_md_solinux.c
```diff
@@ -380,6 +386,13 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                                            asked for? Current model is
                                            fine unless another model
                                            is asked for */
+#ifdef AIX
+    const char *mallocOptionsName = "MALLOCOPTIONS";
+    const char *mallocOptionsValue = "multiheap,buckets";
+    if (setenv(mallocOptionsName, mallocOptionsValue, 0) != 0) {
+      printf("mallocOptins failed\n");
+    }
```
keithc-ca: The types of the variables have not been updated.
The indentation is not fixed: the `printf` line is not properly indented.
There's still a typo in the message, but `printf` is not an appropriate response to a failure of `setenv`.

pshipton: @keithc-ca what are you thinking should be an appropriate response to a failure? It should never occur, unless you are out of environment space or memory?, but if it does it would be nice to have the message that indicate the JVM will perform differently.

keithc-ca: If the consensus is that a message is the appropriate response, then, IMO:
* it should be written to `stderr`
* it should speak to the consequences of the failure, e.g. performance may be affected

shanchao95:
@pshipton  @keithc-ca Thanks. The code is updated.

jdk/src/solaris/bin/java_md_solinux.c
```diff
@@ -380,6 +386,14 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                                            asked for? Current model is
                                            fine unless another model
                                            is asked for */
+#ifdef AIX
+    const char *mallocOptionsName = "MALLOCOPTIONS";
+    const char *mallocOptionsValue = "multiheap,buckets";
+    if (setenv(mallocOptionsName, mallocOptionsValue, 0) != 0) {
+      fprintf(stderr, "mallocOption failed: performance may be affected\n");
+      fflush(stderr);
```
keithc-ca: Please fix the indentation (8 spaces here).
`fflush` is not needed for `stderr`.
I suggest the message should have more information: e.g.
> setenv('MALLOCOPTIONS=multiheap,buckets') failed: performance may be affected

shanchao95: sure, updated

keithc-ca:
This looks good. Can you put pull requests together for jdk11, jdk14 and jdk(next) also? Thanks.

shanchao95:
@keithc-ca Sure, PR is open for jdk11, jdk14 and jdk(next).

keithc-ca:
Jenkins test sanity aix jdk8

keithc-ca:
Jenkins test sanity aix jdk8

keithc-ca:
The build passed even though that fact was not reported here: merging.

PR #367, opened 2020-01-28
Use codesign as appropriate on OSX

keithc-ca:
Sign files on OSX when the code-signing identity specified via configure option `--with-macosx-codesign-identity` is available.

This is proposed as part of the fix for eclipse/openj9#8389.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -84,51 +84,68 @@ endif
 	stage_openj9_build_jdk \
 	#
 
-# jitserver
+define copy_and_sign
+	$(call install-file)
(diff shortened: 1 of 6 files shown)


keithc-ca:
FYI @pshipton

keithc-ca:
Jenkins test sanity osx,win,xlinux,xlinuxcm jdk8

keithc-ca:
Updated to actually use the identify specified via `--with-macosx-codesign-identity`.

keithc-ca:
I managed to do a local build on Catalina(!) only to discover that `OpenJ9.gmk` will need some small changes to avoid attempting to re-sign files.

keithc-ca:
Re-signing problem is fixed.

pshipton:
Jenkins compile osx,win,plinux,xlinuxcm jdk8

PR #371, opened 2020-01-31
Sign bundled crypto libraries

keithc-ca:
When libcrypto.dylib or libssl.dylib are bundled in a jdk on OSX, they must also be signed.

* define CodesignFile in custom-spec.gmk for use in OpenJ9.gmk and CopyFiles.gmk.
* update code signing:
  - copy entitlements.plist from jdk11
  - also sign all shared libraries

Issue: eclipse/openj9#8389

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -86,7 +86,7 @@ endif
 
 define copy_and_sign
 	$(call install-file)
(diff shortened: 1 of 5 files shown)


keithc-ca:
Jenkins test sanity aix,osx,plinux,xlinux,zlinux,win,win32 jdk8

pshipton:
jenkins compile xlinuxcm jdk8

PR #377, opened 2020-02-21
OpenJ9 management and mangement_ext are obsolete

pshipton:
The management shared library is provided by OpenJDK. Remove unnecessary natives, since most of the management native are provided by OpenJ9 in a different shared library. Only FileSystemImpl natives are required.

Depends on https://github.com/eclipse/openj9/pull/8631


pshipton:
Squashed.

PR #380, opened 2020-02-26
Add debug-image target

keithc-ca:
* copy debuginfo files for openj9 shared libraries
* collect files (*.debuginfo, *.pdb, etc.) into images/debug-image

Note: that the default configure behavior is `--enable-zip-debug-info` which results in `*.diz` files being included in the debug image. To get uncompressed debug information files, use `--disable-zip-debug-info` (which doesn't work on Windows due to some missing rules in openjdk makefiles).

### DIFF
--- a/closed/CopyToBuildJdk.gmk
+++ b/closed/CopyToBuildJdk.gmk
@@ -35,6 +35,83 @@ else
   OPENJ9_LIBS_DIR := $(JDK_OUTPUTDIR)/lib$(OPENJDK_TARGET_CPU_LIBDIR)
 endif # windows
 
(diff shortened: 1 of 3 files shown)


keithc-ca:
Jenkins test sanity aix,osx,plinux,win,win32 jdk8

keithc-ca:
Marking as 'WIP' until I correct the paths for `.dSYM` folders on OSX.

keithc-ca:
The proper paths are now computed for all platforms.

pshipton:
jenkins compile osx jdk8

pshipton:
Jenkins test sanity aix,osx,plinux,win,win32 jdk8

PR #382, opened 2020-03-06
Omit inappropriate man pages

keithc-ca:
Don't include man pages:
* for launchers not implemented by OpenJ9
* that describe behavior that differs from the OpenJ9 implementation

See eclipse/openj9#8732.


pshipton:
jenkins compile plinux jdk8

PR #384, opened 2020-03-13
Enable use of '--disable-zip-debug-info' on Windows

keithc-ca:
Add rules so make can see how `.map` and `.pdb` files are created (an exemption is needed for `unpack.dll`).


pshipton:
jenkins test sanity win,win32 jdk8

PR #385, opened 2020-03-13
(0.20.0) Enable use of '--disable-zip-debug-info' on Windows

keithc-ca:
Add rules so make can see how .map and .pdb files are created (an exemption is needed for unpack.dll).

This is the same commit behind #384.


pshipton:
The copyright check failure is for files other than the one being changed by this PR.

PR #388, opened 2020-03-19
Fix handling of '--with-cmake' configure option

keithc-ca:
* 'no' disables use of cmake (the user said `--with-cmake=no` or `--without-cmake`)
* blank or 'yes' enables cmake with the default program (cmake)
* any other value is expected to name the cmake program (absolute or PATH-relative)


keithc-ca:
Jenkins test sanity plinux,xlinuxcm jdk8

keithc-ca:
Jenkins test sanity plinux,win32,xlinuxcm jdk8

PR #391, opened 2020-03-20
Stop checking for protoc

keithc-ca:
* it is no longer required for jitserver

### DIFF
--- a/jdk/make/closed/autoconf/custom-hook.m4
+++ b/jdk/make/closed/autoconf/custom-hook.m4
@@ -218,42 +218,18 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
 
 AC_DEFUN([OPENJ9_CONFIGURE_JITSERVER],
 [
-  AC_MSG_CHECKING([for jitserver])
   AC_ARG_ENABLE([jitserver], [AS_HELP_STRING([--enable-jitserver], [enable JITServer support @<:@disabled@:>@])])
+
+  AC_MSG_CHECKING([for jitserver])
   OPENJ9_ENABLE_JITSERVER=false
-
   if test "x$enable_jitserver" = xyes ; then
-    AC_MSG_RESULT([yes (explicitly enabled)])
-
-    if test "x$OPENJDK_TARGET_OS" != xlinux ; then
-      AC_MSG_ERROR([jitserver is unsupported for $OPENJDK_TARGET_OS])
+    if test "x$OPENJDK_TARGET_OS" = xlinux ; then
+      AC_MSG_RESULT([yes (explicitly enabled)])
+      OPENJ9_ENABLE_JITSERVER=true
     else
-      AC_CHECK_PROG(PROTOC_INSTALLED,protoc,yes,no)
-      if test "x$PROTOC_INSTALLED" = xno ; then
-        AC_MSG_ERROR([jitserver requires protoc])
-      else
-        AC_MSG_CHECKING([protobuf version])
-        if test "x$OPENJ9_CPU" = xx86-64 ; then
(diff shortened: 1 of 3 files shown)


jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -218,42 +218,18 @@ AC_DEFUN([OPENJ9_PLATFORM_EXTRACT_VARS_FROM_CPU],
...
-  OPENJ9_ENABLE_JITSERVER=false
 
+  AC_MSG_CHECKING([for jitserver])
+  OPENJ9_ENABLE_JITSERVER=false
   if test "x$enable_jitserver" = xyes ; then
-    AC_MSG_RESULT([yes (explicitly enabled)])
-
-    if test "x$OPENJDK_TARGET_OS" != xlinux ; then
-      AC_MSG_ERROR([jitserver is unsupported for $OPENJDK_TARGET_OS])
+    if test "x$OPENJDK_TARGET_OS" = xlinux ; then
```
pshipton: I don't understand this, we're building JITServer for plinux.
https://github.com/eclipse/openj9/blob/be03cea0cb7cdd829a61b283d6558b02c542b308/buildenv/jenkins/variables/defaults.yml#L167

pshipton:
@mpirvu pls review this, and the others.
ibmruntimes/openj9-openjdk-jdk11#281
ibmruntimes/openj9-openjdk-jdk14#31
ibmruntimes/openj9-openjdk-jdk#182

jdk/make/closed/autoconf/custom-hook.m4
mpirvu: I have the same question, both xlinux and plinux should be supported

keithc-ca: The `if` was inverted.

jdk/make/closed/autoconf/generated-configure.sh
```diff
@@ -16079,96 +16018,25 @@ $as_echo "no (default for $OPENJ9_PLATFORM_CODE)" >&6; }
...
 
-  OPENJ9_ENABLE_JITSERVER=false
 
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for jitserver" >&5
+$as_echo_n "checking for jitserver... " >&6; }
+  OPENJ9_ENABLE_JITSERVER=false
   if test "x$enable_jitserver" = xyes ; then
-    { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes (explicitly enabled)" >&5
+    if test "x$OPENJDK_TARGET_OS" = xlinux ; then
+      { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes (explicitly enabled)" >&5
```
mpirvu: similarly here, JITServer is supported on both plinux and xlinux.

jdk/make/closed/autoconf/custom-hook.m4
keithc-ca: The `x` in `"x$OPENJDK_TARGET_OS"` is the idiom to safely handle the case where `$OPENJDK_TARGET_OS` is the empty string. If we didn't have to be concerned about that, the equivalent expression would be: `"$OPENJDK_TARGET_OS" = linux`.

jdk/make/closed/autoconf/generated-configure.sh
keithc-ca: See above (this is a derived file).

jdk/make/closed/autoconf/custom-hook.m4
mpirvu: got it.

keithc-ca:
The definition of the `OPENJ9_CONFIGURE_JITSERVER` macro is identical in the related PRs.

PR #399, opened 2020-04-13
CMake: remove VERBOSE from the environment

keithc-ca:
Both cmake and the makefiles it generates are sensitive to the `VERBOSE` environment variable. This removes `VERBOSE` from the VM build environment altogether for build behavior that's consistent with newer jdk versions.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -307,6 +307,11 @@ else
   CMAKE_ARGS += -DJ9VM_OPT_JITSERVER=OFF
 endif # OPENJ9_ENABLE_JITSERVER
 
+ifneq (,$(filter debug trace, $(LOG_LEVEL)))
+  # The user said LOG=debug or LOG=trace, so tell cmake to echo commands.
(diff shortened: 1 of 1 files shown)


keithc-ca:
Jenkins test sanity plinux,xlinuxcm jdk8

pshipton:
Why is VERBOSE set the first place?
What happens if somebody wants VERBOSE output?

keithc-ca:
> Why is VERBOSE set the first place?

VERBOSE is defined and used by jdk8 makefiles.

> What happens if somebody wants VERBOSE output?

I'll add this which will propagate the meaning of `LOG=debug` as defined by openjdk:
```
ifneq (,$(filter debug trace, $(LOG_LEVEL)))
  CMAKE_ARGS += -DCMAKE_VERBOSE_MAKEFILE=ON
endif
```

keithc-ca:
Updated for https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/399#issuecomment-613101441.

pshipton:
@dnakamura pls review

pshipton:
jenkins compile xlinuxcm jdk8

PR #427, opened 2020-07-07
Reduce differences relative to openjdk base

keithc-ca:
We no longer need to comment out the code that is now conditional due to:
* 8248399: Build installs jfr binary when JFR is disabled


pshipton:
jenkins compile plinux jdk8

PR #437, opened 2020-07-28
Change openjdk-tag and version-number for jdk8u265-b01 update

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>


lumpfish:
Copyright check failed because the openjdk-tag file has no copyright (which is correct).

PR #441, opened 2020-08-18
Add OSX get-task-allow entitlement

keithc-ca:
Issue: AdoptOpenJDK/openjdk-build#1778


DanHeidinga:
Merging without a build as this will only be tested by the notarization process

PR #444, opened 2020-09-18
For easier reading, use OpenJ9 text instead of html license

pshipton:
Also include epl license which the OpenJ9 license refers to.

Similar change to https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/344/


pshipton:
jenkins compile zlinux jdk8

pshipton:
jenkins compile zlinux jdk8

keithc-ca (approved):

PR #446, opened 2020-09-30
Update the openj9-0.23.0 release branch with the jdk8u272-b09 content 

pshipton:


keithc-ca (approved):
Ideally there would be a better comment in the 'Merge commit', but I won't nitpick.

pshipton:
I think if I create an appropriately named branch at the merge point the generated message will refer to it instead of the SHA. I'll try that next time.

keithc-ca:
I use the 'Merge...' in the context menu in git repositories view in eclipse. It allows you to perform the merge without committing it, giving you a chance to edit the commit message. From the command line, `git merge --no-commit ...` is what you want.

PR #449, opened 2020-10-15
Update JDK8 windows compiles to use Visual Studio 2013

jdekonin:
* [skip ci]
* disable the use of rebase
* eclipse/openj9#10129

Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -148,6 +148,9 @@ else
   FEATURE_SED_SCRIPT += $(call SedDisable,opt_jitserver)
 endif
 
+# Disable windows rebase.
+SPEC_SED_SCRIPT += $(call SedDisable,uma_windowsRebase)
+
 # Adjust OpenJDK MethodHandles enablement flags.
 ifeq (true,$(OPENJ9_ENABLE_OPENJDK_METHODHANDLES))
   FEATURE_SED_SCRIPT += $(call SedEnable,opt_openjdkMethodhandle)
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4410,7 +4410,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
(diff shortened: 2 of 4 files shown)


jdekonin:
Needs to merge with https://github.com/eclipse/openj9/pull/10892

closed/OpenJ9.gmk
```diff
@@ -378,6 +381,7 @@ run-preprocessors-j9 : stage-j9
 			JAVA_HOME=$(BOOT_JDK) \
 			JAVA_VERSION=80 \
 			OMR_DIR=$(OPENJ9_VM_BUILD_DIR)/omr \
+			OPENJ9_BUILD=true \
```
keithc-ca, line 384: Why is this added? I don't see any dependency on that macro in `buildtools.mk`.

common/autoconf/generated-configure.sh
```diff
@@ -1014,7 +1014,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
```
keithc-ca, line 1017: Can you regenerate on a newer system so we don't get these unrelated changes?

keithc-ca (changes requested):

closed/OpenJ9.gmk
jdekonin: I thought I had traced this back to a previous PR as being required, but retesting now show that its not needed.

common/autoconf/generated-configure.sh
jdekonin: Done

keithc-ca (approved):

keithc-ca:
@pshipton I'll let you merge this and eclipse/openj9#10892 when you get satisfactory answers from Joe.

pshipton (approved):

pshipton:
Hmm, I'm thinking that once this is merged all builds need to use VS2013 or they will fail. That include builds at Adopt. https://github.com/AdoptOpenJDK/openjdk-build/issues/466#issuecomment-717348184

pshipton:
@jdekonin there is a merge conflict here.

jdekonin:
Rebase performed and conflict resolved @pshipton

PR #453, opened 2020-10-22
Improve integration with openjdk build system

keithc-ca:
This is a back-port of much of ibmruntimes/openj9-openjdk-jdk11#355 for Java 8, but, unfortunately, I was unable to propagate the job limit to cmake.


pshipton:
jenkins compile aix,win,plinux,osx jdk8

pshipton (approved):

PR #458, opened 2020-11-12
Add new configure option --with-mixedrefs for mixed pointer mode

sharon-wang:
JDK8 port of https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/359.

Related Issue: https://github.com/eclipse/openj9/issues/8878

PR #459, opened 2020-11-16
Add check for OPENJDK_TARGET_CPU_BITS back into configure

sharon-wang:
In the changes from https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/458, a check for `"x$OPENJDK_TARGET_CPU_BITS" = x32` was mistakenly removed. This resulted in the build getting generated in the `compressedrefs` directory, instead of the `default` directory, causing the tests in https://github.com/eclipse/openj9/issues/11187 to fail. Adding this check back in should fix https://github.com/eclipse/openj9/issues/11187.


keithc-ca (approved):
When we address ibmruntimes/openj9-openjdk-jdk11#361 we should add more checking (e.g. mixedrefs should not be accepted for (supported) 32-bit targets).

PR #73, opened 2018-04-20
Include msvcp100.dll in Windows builds

pshipton:
Fixes: eclipse/openj9#1632

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>

PR #460, opened 2020-11-16
Remove build redundancies relating to msvcp.dll

keithc-ca:
#73 added `msvcpXXX.dll` in several places, including `bin` and `jre/bin` unlike a Hotspot JDK which only includes the latter.

This removes the duplicate code which finds the proper version of `msvcpXXX.dll` and places it in the desired places in a jdk or a jre image.

### DIFF
--- a/closed/make/CopyFiles.gmk
+++ b/closed/make/CopyFiles.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2020 All Rights Reserved
 # ===========================================================================
 # 
 # This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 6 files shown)


keithc-ca:
Jenkins test sanity win,win32 jdk8

pshipton:
Do I understand correctly that this change removes `msvcpXXX.dll` from bin? This scares me a little, but I can't find any reason why it would be needed there, there is nothing in the bin directory that uses `msvcpXXX.dll` directly.

pshipton:
In case there is something using the invocation API that will break.

jenkins test extended win,win32 jdk8

keithc-ca:
> Do I understand correctly that this change removes `msvcpXXX.dll` from bin? This scares me a little, but I can't find any reason why it would be needed there, there is nothing in the bin directory that uses `msvcpXXX.dll` directly.

Yes, that is correct. I was a little concerned at first myself, but as you can see, all the sanity testing passed.

pshipton (approved):

PR #462, opened 2020-11-17
Set JAVA_VENDOR and JAVA_VENDOR_URL defines

pshipton:
Set OpenJ9 defaults for COMPANY_NAME and VENDOR_URL.


closed/OpenJ9.gmk
```diff
@@ -203,6 +203,8 @@ OPENJ9_VERSION_VARS := \
 	OPENJDK_TARGET_OS \
 	PRODUCT_NAME \
 	USERNAME \
+	COMPANY_NAME \
+	VENDOR_URL \
```
keithc-ca, line 207: Please retain these in sorted order.

common/autoconf/jdk-options.m4
```diff
@@ -568,7 +569,9 @@ AC_DEFUN_ONCE([JDKOPT_SETUP_JDK_VERSION_NUMBERS],
     AC_MSG_ERROR([--with-vendor-url must have a value])
   elif [ ! [[ $with_vendor_url =~ ^[[:print:]]*$ ]] ]; then
     AC_MSG_ERROR([--with-vendor-url contains non-printing characters: $with_vendor_url])
-  else
+  elif test "x$with_vendor_url" != x; then
```
keithc-ca, line 572: This doesn't handle the use of `--without-vendor-url`.

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -380,9 +380,6 @@ AC_DEFUN([OPENJ9_CHECK_NASM_VERSION],
 
 AC_DEFUN([OPENJDK_VERSION_DETAILS],
 [
-  # Source the closed version numbers
-  . $SRC_ROOT/jdk/make/closed/autoconf/openj9ext-version-numbers
```
keithc-ca, line 384: Why does this need to move? Isn't `CUSTOM_EARLY_HOOK` early enough?

pshipton: Correct, `CUSTOM_EARLY_HOOK` is not early enough.

common/autoconf/jdk-options.m4
pshipton: Such a thing is not supported afaik.

closed/OpenJ9.gmk
pshipton: Updated.

jdk/make/closed/autoconf/custom-hook.m4
pshipton: Besides not being early enough, not doing it here is just problematic. The values are contained in the generated spec.gmk file, and if it's not done here spec.gmk will have the wrong values, which could get used when building various parts of the JDK. i.e. even if setting COMPANY_NAME later is used for some part of the OpenJ9 build, there are likely other parts using the wrong value.

pshipton: CUSTOM_EARLY_HOOK is run before JDKOPT_SETUP_JDK_VERSION_NUMBERS, but JDKOPT_SETUP_JDK_VERSION_NUMBERS runs `$AUTOCONF_DIR/version-numbers` which would overwrite any values set earlier.

pshipton: I've done some editing of the above comments, best to read the latest versions.

common/autoconf/jdk-options.m4
keithc-ca: Here's an excerpt from `bash configure --help`:
```
Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
```
(in this case, 'PACKAGE' is 'vendor-url').

pshipton: Setting `--with-vendor-url=yes` is an error. I think this is correct, and duplicates how it works for jdk11+.

keithc-ca: The last line I quoted was the one that is relevant here; the code should be:
```
elif test "x$with_vendor_url" != x -a "x$with_vendor_url" != xno ; then
```

keithc-ca: Actually, that's not what a user might expect; if they say `--without-vendor-url`, I think they should get no URL, not the default, so we want something like this:
```
elif test "x$with_vendor_url" = xno ; then
  VENDOR_URL=
elif test "x$with_vendor_url" != x ; then
  VENDOR_URL="$with_vendor_url"
```

pshipton: That's no how any of the other settings work, or how jdk11+ works.

keithc-ca: See:
* https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/common/autoconf/basics.m4#L481
* https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/common/autoconf/toolchain.m4#L857

pshipton: By "other settings" I mean `--with-vendor-bug-url`, `--with-vendor-vm-bug-url`, `--with-vendor-name` (possibly more). I don't think we should have `--with_vendor_url` work differently from these other options or from jdk11+.

keithc-ca: Running configure without `--with-vendor-url` today yields
```
VENDOR_URL:=
```
in `spec.gmk`; with this change, there is no way to achieve that.
With my suggestion, `--without-vendor-url` would leave it blank.

pshipton: jdk11+ doesn't allow setting VENDOR_URL blank either. I don't think we should support having it blank, it would affect the system properties set by OpenJ9.

keithc-ca: This excerpt from jdk11's `spec.gmk.in` would seem to suggest otherwise
```
# Only export VENDOR_URL, VENDOR_URL_BUG and VENDOR_VM_URL_BUG to the build if
# they are not empty. Otherwise, default values which are defined in the sources
# will be used.
ifneq ($(VENDOR_URL),)
  VERSION_CFLAGS += -DVENDOR_URL='"$(VENDOR_URL)"'
endif
```
but it appears you're right, `VENDOR_URL` can't be blank there.

keithc-ca (approved):

keithc-ca:
Jenkins compile zlinux jdk8 depends eclipse/openj9#11216

keithc-ca:
I asked @AdamBrousseau to investigate errors like this:
```
13:17:46  stderr: fatal: unable to access 'https://github.com/eclipse/openj9.git/': Could not resolve host: github.com
```

pshipton:
See https://openj9.slack.com/archives/CDS7QE9HB/p1605723037109300

keithc-ca:
Jenkins compile zlinux jdk8 depends eclipse/openj9#11216

PR #464, opened 2020-11-23
Suppress use of broken make rules inherited from openjdk

keithc-ca:
* the rule to produce `java.dSYM` is broken: the target is expected to be a folder (without `-R`, `rm` & `cp` won't do what's necessary) and we already have rules to produce the files within that folder
* the rule to produce `unpack200.dSYM` is also broken: it copies a file rather than the debug info files one would expect

Fixes: eclipse/openj9#11244.


keithc-ca:
Jenkins compile osx,win32 jdk8

pshipton (approved):

PR #473, opened 2021-01-25
Add accelerated zlibNX to LIBPATH for AIX P9 or newer systems and set system zlib as default on AIX

rmnattas:
`zlibNX` is an accelerated replacement for `zlib`, it make use of the NX accelerator in P9 and newer systems.
This PR enable the use of `zlibNX` for AIX systems by adding it to the `LIBPATH` if the system is P9 or newer and has the NX accelerator enabled.

This PR also sets system zlib as default instead of bundled zlib on AIX systems to allow `zlibNX`

Signed-off-by: Abdulrahman Alattas <rmnattas@gmail.com>

### DIFF
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4031,6 +4031,10 @@ fi
 # questions.
 #
 
+# ===========================================================================
+# (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+# ===========================================================================
+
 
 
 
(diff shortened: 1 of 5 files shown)


rmnattas:
@pshipton @keithc-ca @zl-wang

pshipton:
I think we should mention at least this jdk8 change in behavior in the User Guide. Please create a docs issue at https://github.com/eclipse/openj9-docs and preferably a PR as well to update the release notes for the next release.

rmnattas:
> I think we should mention at least this jdk8 change in behavior in the User Guide. Please create a docs issue at https://github.com/eclipse/openj9-docs and preferably a PR as well to update the release notes for the next release.

https://github.com/eclipse/openj9-docs/issues/722

keithc-ca:
Jenkins test sanity aix,xlinux jdk8

keithc-ca (approved):

PR #476, opened 2021-02-03
Merging latest openj9 into openj9-staging - Update OPENJDK_TAG to jdk8u292-b01

JasonFengJ9:
Resolved merging conflicts:

```
  jdk/src/solaris/bin/aarch64/jvm.cfg
  jdk/src/share/native/com/sun/media/sound/SoundDefs.h
  jdk/make/lib/SoundLibraries.gmk
  common/autoconf/generated-configure.sh
```


pshipton:
I note there are some changes we could make to help avoid future merge conflicts, although it doesn't need to occur in this PR.

pshipton:
lgtm

keithc-ca:
The title says this merges from `jdk8u292-b00`, but that tag refers to content that is already in the openj9 branch.
Is the intent here to merge with `jdk8u292-b01`?

JasonFengJ9:
> but that tag refers to content that is already in the openj9 branch.
Is the intent here to merge with jdk8u292-b01?

It intended to merge with `jdk8u292-b00`.
What content tagged is already in the `openj9` branch?

keithc-ca:
The tag 'jdk8u292-b00' points to de7aa62cfa02b70641093496899e1abe4e7b05a1 which is an ancestor commit of a1b8db283e3775be94350088a00b8df072eb26bb which is where the openj9 is at currently.

closed/openjdk-tag.gmk
```diff
@@ -1 +1 @@
-OPENJDK_TAG := jdk8u282-b08
+OPENJDK_TAG := jdk8u292-b00
```
keithc-ca, line 1: This should be `jdk8u292-b01`.

common/autoconf/jdk-options.m4
```diff
@@ -165,9 +165,6 @@ AC_DEFUN_ONCE([JDKOPT_SETUP_JVM_VARIANTS],
   if test "x$VAR_CPU" = xppc64 -o "x$VAR_CPU" = xppc64le ; then
     INCLUDE_SA=false
   fi
-  if test "x$OPENJDK_TARGET_CPU" = xaarch64; then
-    INCLUDE_SA=false
-  fi
```
keithc-ca, line 170: I don't think we want this change.

jdk/make/lib/SoundLibraries.gmk
```diff
@@ -152,9 +152,8 @@ else
   endif
 
   ifeq ($(OPENJDK_TARGET_CPU), aarch64)
-    LIBJSOUND_CFLAGS += -DX_ARCH=X_AARCH64
+	LIBJSOUND_CFLAGS += -DX_ARCH=X_AARCH64
```
keithc-ca, line 155: We don't really want this change: the leading tab could be problematic.

keithc-ca (changes requested):
I'd like to see SoundDefs.h change so `X_AARCH64` is matches openjdk.

common/autoconf/jdk-options.m4
JasonFengJ9: This is from OpenJDK, is there an issue/pr that requires the code snippet being deleted?

jdk/make/lib/SoundLibraries.gmk
JasonFengJ9: sure, we can keep initial JCL patch instead.

closed/openjdk-tag.gmk
JasonFengJ9: agreed

common/autoconf/jdk-options.m4
keithc-ca: The change is to delete those lines. We don't want that; we want to continue to support aarch64, so those lines should stay.

JasonFengJ9: As per https://bugs.openjdk.java.net/browse/JDK-8155627, this change is to _Enable SA on AArch64_ 
        _The serviceability agent is currently disabled. There is no good reason for this._
I think OpenJ9 can use this change to enable SA.

keithc-ca: If we think SA should be enabled, it should be done in a separate pull request specifically for that purpose.

JasonFengJ9: This commit comes in with OpenJDK updates as part of merging process, if a change to this OpenJDK commit (i.e., JCL patch) is required, it should be in a separated PR instead, does this make sense?

keithc-ca: I misunderstood the source of those lines. They originated with openjdk and were not changed for openj9. We should follow openjdk and remove them too. Sorry for the diversion.

JasonFengJ9:
Updated the PR, please have another look.

keithc-ca:
`jdk/make/closed/autoconf/generated-configure.sh` needs to be regenerated to reflect changes to `common/autoconf/jdk-options.m4`.

JasonFengJ9:
> jdk/make/closed/autoconf/generated-configure.sh needs to be regenerated to reflect changes to common/autoconf/jdk-options.m4.

Re-generated `generated-configure.sh` via `common/autoconf/autogen.sh` https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/476/commits/8cb1be3c5035acc3ffc905bbc976a9e79c1cd615.

keithc-ca:
Please redo `common/autoconf/autogen.sh` and include both generated files. They each have an embedded timestamp which I believe need to be consistent.

keithc-ca (approved):

keithc-ca:
Jenkins test sanity aix,osx,plinux,win32 jdk8

pshipton:
The Windows failure is https://github.com/eclipse/openj9/pull/11869

pshipton:
Note this could just be merged, as it will through the acceptance build anyway. The only downside is possibly having to open another PR to fix problems.

keithc-ca:
> Note this could just be merged, as it will go through the acceptance build anyway.

That seems reasonable. Please approve if you're happy and I will merge.

pshipton (approved):

PR #490, opened 2021-03-22
Support including healthcenter agent

keithc-ca:
* unpack artifacts from archive named via `--with-healthcenter` option
* set default transport to `jrmp`

This is similar to ibmruntimes/openj9-openjdk-jdk11#395 but for Java 8.


JasonFengJ9:
> unpack artifacts from archive named via --with-healthcenter option

Is this PR a completed solution to integrate `healthcenter` including its native library as well?
A sanity test can be introduced, which might check if the `healthcenter` presents, and run `-Xhealthcenter -version`.

pshipton:
jenkins compile all jdk8

keithc-ca:
Yes, wherever openj9 is built with healthcenter, that sound like a reasonable sanity test, however, I don't know that there are plans to create such builds regularly.

pshipton:
True for the open builds. z/OS should be regularly including it. It seems the test could be in the open but only enabled for z/OS (when it becomes applicable).

pshipton (approved):

PR #502, opened 2021-05-20
In test, increase deflate buffer size for zlinux Ubuntu 20

pshipton:
jdk/test/java/util/zip/DeInflate.java

Issue eclipse-openj9/openj9#12740


pshipton:
This change is not needed on jdk8, the test passes without it.

PR #506, opened 2021-06-09
DDR: generate virtual structure stub classes

keithc-ca:
Stub classes are needed even for those types which no longer exist but have optional fields.
Use new `-a` option of `StructureStubGenerator`.

Depends on eclipse-openj9/openj9#12908.


pshipton:
jenkins compile zlinux,aix,win,osx jdk8

pshipton (approved):

PR #507, opened 2021-06-10
Handle Exceptional Padding Cases in the Native RSADP Verification

samasri:
This PR fixes part of eclipse-openj9/openj9#11500 by changing the way verification is implemented in the RSADP operation.

More specifically, this PR adds a special case when verifying a message. For certain key sizes (where keySize % 8 == 1), the decrypted message retrieved from the `RSA_public_decrypt` includes a 1 byte padding at the beginning of the message. In these cases, this padding must be ignored when comparing the decrypted message to the original message.

Signed-off-by: Samer AL Masri <samasri@ibm.com>


keithc-ca:
Jenkins compile win32 jdk8

keithc-ca (approved):

keithc-ca:
This is the same change to the same function as in ibmruntimes/openj9-openjdk-jdk11#404: a compile should be sufficient validation.

PR #518, opened 2021-06-30
Set default load address for launchers on plinux

pshipton:
On some machines the java launcher is loaded around 0x100000000,
breaking up the address space.

Once this is reviewed I'll create the same change for the other versions.

### DIFF
--- a/jdk/make/CompileLaunchers.gmk
+++ b/jdk/make/CompileLaunchers.gmk
@@ -94,10 +94,11 @@ define SetupLauncher
   $1_LDFLAGS := $3
 
   ifeq ($(OPENJDK_TARGET_OS), linux)
-  # Set the image base address for zLinux 64 to 0x60000 for launchers,
-  # allows compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
-  # / RTC PR 100052
-    ifeq ($(OPENJDK_TARGET_CPU), s390x)
+    # Set the image base address for p/zLinux 64 to 0x60000 for launchers,
+    # allows compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
+    # Also maximizes the address space available for the java heap.
+    # / RTC PR 100052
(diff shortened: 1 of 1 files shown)


jdk/make/CompileLaunchers.gmk
```diff
@@ -94,12 +94,16 @@ define SetupLauncher
...
   # allows compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
+  # Also maximizes the address space available for the java heap.
   # / RTC PR 100052
     ifeq ($(OPENJDK_TARGET_CPU), s390x)
       $1_LDFLAGS += -Wl,-Ttext-segment=0x60000
     endif
+    ifeq ($(OPENJDK_TARGET_CPU), ppc64le)
+      $1_LDFLAGS += -Wl,-Ttext-segment=0x60000
+    endif
   endif
```
keithc-ca, line 107: The comments should be indented the same as the rest of the body of `ifeq`.
Since we want to treat p and z the same, the cases can be merged.
```
  ifeq ($(OPENJDK_TARGET_OS), linux)
    # Set the image base address for p/zLinux 64 to 0x60000 for launchers,
    # allows compressedRefsShift to be 0 when -Xmx is set to 2040m or more.
    # Also maximizes the address space available for the java heap.
    # /  RTC PR 100052
    ifneq (,$(filter ppc64le s390x, $(OPENJDK_TARGET_CPU)))
      $1_LDFLAGS += -Wl,-Ttext-segment=0x60000
    endif
  endif
```

keithc-ca (approved):

keithc-ca:
Jenkins test sanity plinux,xlinux,zlinux jdk8

keithc-ca:
Jenkins test sanity xlinux jdk8

keithc-ca:
The xLinux failure is an infrastructure problem.

pshipton:
For the record, I built xlinux with this change and checked it manually, it's not affected by this change as expected.
```
Entry point address:               0x654
```

PR #527, opened 2021-08-06
Merge master jdk8u312-b01 into openj9-staging

JasonFengJ9:
Resolved conflicts:
```
	common/autoconf/generated-configure.sh
```
Re-generated via `common/autoconf/autogen.sh`.
Note: `autoconf` in use is `2.71` (previously generated by `2.69`).


keithc-ca:
Please re-run autogen.sh using autoconf 2.69; if we decide to upgrade to 2.71 that should be done separately.

JasonFengJ9:
Re-run autogen.sh w/ autoconf 2.69, please have another look.

keithc-ca:
That looks better.

In the future, I think the process should merge with the openj9 branch as the first step (as opposed to the last). In this instance that would have been a fast-forward merge (I expect probably most often that would be the case); and you would not have had to merge the generated scripts twice.

keithc-ca (approved):

JasonFengJ9:
> In the future, I think the process should merge with the openj9 branch as the first step (as opposed to the last). 

This change is required at `merge_master_to_staging.sh` first if we proceed that order.

keithc-ca:
> This change is required at `merge_master_to_staging.sh` first if we proceed that order.

Can you please look into that?

JasonFengJ9:
Sure, putting into my backlog.

PR #528, opened 2021-08-09
Exit immediately if any unexpected error occurs cloning source repos

keithc-ca:
As is done for other jdk versions, e.g. https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/get_source.sh#L26


pshipton:
jenkins compile all jdk8

pshipton (approved):

PR #530, opened 2021-08-30
Ignore transient negative pool.ctl in getSurplusQueuedTaskCount()

keithc-ca:
See explanation in https://github.com/eclipse-openj9/openj9/issues/12073#issuecomment-908539871.


keithc-ca:
Either the test must change or the fork-join change will be different.

PR #542, opened 2022-01-17
Set native stack size using -Xmso instead of -Xss

mikezhang1234567890:
OpenJ9 uses -Xmso to set native stack size, so launcher should also use this
value when creating the JVM thread.

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

Port of ibmruntimes/openj9-openjdk-jdk17#57 to JDK8
Partially addresses eclipse-openj9/openj9#12785

### DIFF
--- a/jdk/src/share/bin/java.c
+++ b/jdk/src/share/bin/java.c
@@ -24,6 +24,12 @@
  */
 
 /*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


keithc-ca (approved):

keithc-ca:
jenkins test sanity zlinux jdk8

PR #543, opened 2022-01-21
Support configure option '--with-openssl=fetched' on Windows

keithc-ca:
Issue: eclipse-openj9/openj9#14302.


closed/openssl.gmk
```diff
@@ -18,50 +18,65 @@
...
+  ifeq ($(OPENJDK_TARGET_CPU), ppc64le)
+    OPENSSL_TARGET := linux-ppc64le
+  else ifeq ($(OPENJDK_TARGET_CPU), s390x)
+    OPENSSL_TARGET := linux64-s390x
+  else ifeq ($(OPENJDK_TARGET_CPU), x86_64)
+    OPENSSL_TARGET := linux-x86_64
+  endif
+else ifeq ($(OPENJDK_TARGET_OS), macosx)
+  ifeq ($(OPENJDK_TARGET_CPU), arm64)
+    OPENSSL_TARGET := darwin64-x86_64-cc
```
pshipton, line 64: This specifies x86 instead of arm.

keithc-ca: Good catch. I inverted that 'ifeq' element badly.

keithc-ca: I plan to replace the series of CPU cases to:
```
  ifneq (,$(filter arm64 x86_64, $(OPENJDK_TARGET_CPU)))
    OPENSSL_TARGET := darwin64-$(OPENJDK_TARGET_CPU)-cc
  endif
```
and apply a similar change for Linux.

keithc-ca:
Fixed calculation of `OPENSSL_TARGET`.

keithc-ca:
Updated calculation of `OPENSSL_TARGET`; removed unnecessary include of `MakeBase.gmk`.

pshipton (approved):

PR #546, opened 2022-02-17
Add support for AArch64 macOS in closed/openssl.gmk

keithc-ca:
This is a backport of ibmruntimes/openj9-openjdk-jdk11#476.


pshipton (approved):

knn-k:
It is more complicated to add support for AArch64 macOS to JDK 8 than to do that to newer JDKs.  I have not been successful yet.

PR #553, opened 2022-03-30
Fix @Stable annotation in MethodHandleImpl

keithc-ca:
This is a back-port of ibmruntimes/openj9-openjdk-jdk17#81, required when `--enable-openjdk-methodhandles` is used at configuration time.


keithc-ca:
FYI @knn-k

PR #556, opened 2022-04-05
Remove dead code

keithc-ca:
The code for clang is identical to, and included in, the block above.
This removes a difference relative to upstream.


pshipton:
> This removes a difference relative to upstream.

That's wierd, did we do a merge incorrectly?

jenkins compile mac jdk8

keithc-ca:
> That's weird, did we do a merge incorrectly?

I expect that we just didn't notice that change arriving from upstream.

pshipton (approved):

PR #555, opened 2022-04-04
Compile with -qstackprotect on xlc

pshipton:
Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>

PR #557, opened 2022-04-05
(0.32) Compile with -qstackprotect on xlc

pshipton:
Cherry pick https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/555 for the 0.32 release.


pshipton:
I don't see any differences in the generated-configure files from the head stream, so I think the cherry pick is appropriate.

PR #558, opened 2022-04-08
Temporarily disable CRIU support on JDK8

tajila:
Temporarily disable criu support on JDK8

There are currently efforts to enable CRIU security support to openj9.
The intention is to do this in a phased approach by first support jdk11+
since the code is fairly similar, after JDK8.

Signed-off-by: Tobi Ajila <atobia@ca.ibm.com>

### DIFF
--- a/common/autoconf/generated-configure.sh
+++ b/common/autoconf/generated-configure.sh
@@ -4433,7 +4433,7 @@ VS_SDK_PLATFORM_NAME_2017=
 #CUSTOM_AUTOCONF_INCLUDE
 
 # Do not change or remove the following line, it is needed for consistency checks:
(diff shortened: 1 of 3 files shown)


keithc-ca:
The generated files were not updated properly; run `bash common/autoconf/autogen.sh` to do that.

What besides ibmruntimes/openj9-openjdk-jdk11#498 needs to be back-ported here?
It that's all, it would seem easier to make that single change rather than (at least) three changes.

Is there an issue tracking all the partially completed work (e.g. porting jdk11#498 to other version)?

keithc-ca (approved):

keithc-ca:
jenkins compile osx jdk8

PR #560, opened 2022-04-14
Backport 8259922: MethodHandles.collectArguments does not throw IAE

babsingh:
Fix for https://bugs.openjdk.java.net/browse/JDK-8259922.

Fix: https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/ab6b9c35c7541ea628b028cb1357b904a72de5d9.

### DIFF
--- a/jdk/src/share/classes/java/lang/invoke/MethodHandles.java
+++ b/jdk/src/share/classes/java/lang/invoke/MethodHandles.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -2768,6 +2768,10 @@ assertEquals("[top, [[up, down, strange], charm], bottom]",
         MethodType filterType = filter.type();
         Class<?> rtype = filterType.returnType();
         List<Class<?>> filterArgs = filterType.parameterList();
+        if (pos < 0 || (rtype == void.class && pos > targetType.parameterCount()) ||
+                       (rtype != void.class && pos >= targetType.parameterCount())) {
+            throw newIllegalArgumentException("position is out of range for target", target, pos);
+        }
         if (rtype == void.class) {
             return targetType.insertParameterTypes(pos, filterArgs);
         }


babsingh:
fyi @pshipton Identical to https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/29. For supporting OJDK MHs.

pshipton:
https://github.com/ibmruntimes/openj9-openjdk-jdk16/pull/29/commits/f6e2b25af55d44 also added a test MethodHandlesCollectArgsTest.java, I think we want that as well?

pshipton:
Maybe we'd have to exclude the test until we switch?

pshipton:
Why do we need to backport this if OpenJDK hasn't bothered to backport to 8, 11?

babsingh:
> Why do we need to backport this if OpenJDK hasn't bothered to backport to 8, 11?

Because there are tests in OpenJ9, which will fail without this fix once OJDK MHs are enabled: https://github.com/eclipse-openj9/openj9/issues/11922.

> We're backporting an OpenJDK change, so we should backport the Oracle copyright, not add an IBM copyright.

Updated Oracle copyright and removed IBM copyright.

> https://github.com/ibmruntimes/openj9-openjdk-jdk16/commit/f6e2b25af55d44 also added a test MethodHandlesCollectArgsTest.java, I think we want that as well?

This test won't work with JDK8 because `MethodHandles.empty` is not available in JDK8. We can either update the test or rely on the existing OJ9 tests for coverage in JDK8. For JDK11, I have included the test in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/560. 
 
> Maybe we'd have to exclude the test until we switch?

We do not need to exclude the test because the test should correctly work with OJ9 MHs

pshipton:
jenkins test sanity,extended alinux jdk8

jdk/src/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2022, Oracle and/or its affiliates. All rights reserved.
```
pshipton, line 2: https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/ab6b9c35c7541ea628b028cb1357b904a72de5d9 has a 2021 end date.

pshipton:
Oops, PR testing isn't useful here since the class being modified isn't used atm. Luckily it failed anyway because I got the platform name wrong.

jdk/src/share/classes/java/lang/invoke/MethodHandles.java
babsingh: Updated to match the 2021 end date.

pshipton (approved):

PR #563, opened 2022-04-20
Add openssl version 3+ support for Linux platforms

WilburZjh:
Signed-off-by: Jinhang Zhang <Jinhang.Zhang@ibm.com>

PR #565, opened 2022-04-21
Correct copyright notices

keithc-ca:
Put Oracle copyright notice first, as expected by the automated check.

Addresses the failure seen in #563.

### DIFF
--- a/closed/adds/jdk/src/windows/native/jdk/crypto/jniprovider/NativeCrypto_md.c
+++ b/closed/adds/jdk/src/windows/native/jdk/crypto/jniprovider/NativeCrypto_md.c
@@ -1,14 +1,12 @@
 /*
- * ===========================================================================
- * (c) Copyright IBM Corp. 2019, 2022 All Rights Reserved
- * ===========================================================================
+ * Copyright (c) 1997, 2015, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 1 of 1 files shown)


pshipton:
This file is created by IBM, it shouldn't have any Oracle copyright/license. The automated check is wrong to expect an Oracle copyright.

keithc-ca:
Parts of the file appear to be derived from Oracle code, for example, `JLI_Snprintf()` seems to be a copy from `jdk/src/windows/bin/java_md.c`.

pshipton:
It does look like the GetJREPath() and GetApplicationHome() functions are copied from java_md.c, so I retract my previous statement and agree an Oracle copyright/license is appropriate. Did the copyright checker somehow figure this out?

Other than the copied code, why is this file different from closed/adds/jdk/src/solaris/native/jdk/crypto/jniprovider/NativeCrypto_md.c ?

keithc-ca:
The `solaris` variant only has IBM-authored code.

pshipton:
Ya but, did the copyright checker somehow figure that out, or is configured to expect an Oracle copyright for the windows file?

The commit description needs to be updated, as "as expected by the automated check" is not the proper reason for the change.

keithc-ca:
When both flavours of copyright notices are present, the checker expects the Oracle notice to appear before the IBM notice.

pshipton:
Ok, now I understand. I didn't originally notice the Oracle copyright buried in the middle of the file.

keithc-ca:
I didn't notice it either, but this puts it where it belongs.

PR #569, opened 2022-05-05
Avoid sign extension when calling adler32() and crc32()

pshipton:
adler32() and crc32() functions take unsigned long parameters. Sign
extending the jint causes a crc32() failure with zlib 1.2.12 although it
works with earlier versions. No problem was discovered using adler32()
with zlib 1.2.12 but it should avoid the sign extension as well.

Issue https://github.com/eclipse-openj9/openj9/issues/14953


keithc-ca (approved):

keithc-ca:
jenkins compile alinux64 jdk8

PR #581, opened 2022-06-24
Apply patch to freetype.vcxproj as necessary

keithc-ca:
Update `freetype.vcxproj` from the patch introduced in 16620b953a4d1e4ca91b795b44fe6e44201e4708 when configure option `--with-freetype-src` is used with VS toolchain version 2017 or newer.

Ignore files produced when `--with-freetype-src` is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@
 /openj9
 /openssl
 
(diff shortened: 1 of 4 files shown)


pshipton:
jenkins compile win,win32 jdk8

pshipton (approved):

PR #584, opened 2022-06-29
Improve support for --with-freetype-src

keithc-ca:
Explicitly specify `WindowsTargetPlatformVersion` property so freetype versions 2.5.3, 2.8.1, 2.9.1 or 2.12.1 can be compiled by VS2013 and V2017 (at least).

Remove steps in github actions that patch `freetype.vcxproj` as that is no longer necessary and would be inappropriate for any version of freetype other than 2.8.1.


pshipton:
jenkins compile win,win32 jdk8

pshipton (approved):

PR #601, opened 2022-08-16
FIPS Support for Linux x86 RHEL

WilburZjh:
Signed-off-by: Jinhang Zhang <Jinhang.Zhang@ibm.com>

### DIFF
--- a/closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
+++ b/closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
@@ -1,0 +1,159 @@
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * ===========================================================================
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * IBM designates this particular file as subject to the "Classpath" exception
+ * as provided by IBM in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+package openj9.internal.security;
+
+import java.util.Iterator;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+import sun.security.util.Debug;
+
+/**
+ * Configures the security providers when in FIPS mode.
+ */
+public final class FIPSConfigurator {
+
+    private static final Debug debug = Debug.getInstance("semerufips");
+
+    // FIPS mode enable check, only supported on Linux x64.
+    private static final boolean userEnabledFIPS;
+    private static final boolean isFIPSSupported;
+    private static final boolean shouldEnableFIPS;
+
+    static {
+        String[] props = AccessController.doPrivileged(
+                new PrivilegedAction<String[]>() {
+                    @Override
+                    public String[] run() {
+                        return new String[] {System.getProperty("semeru.fips"),
+                                System.getProperty("os.name"),
+                                System.getProperty("os.arch")};
+                    }
+                });
+        userEnabledFIPS = Boolean.parseBoolean(props[0]);
+        isFIPSSupported = "Linux".equalsIgnoreCase(props[1])
+                && "amd64".equalsIgnoreCase(props[2]);
+        shouldEnableFIPS = userEnabledFIPS && isFIPSSupported;
+    }
+
+    private FIPSConfigurator() {
+        super();
+    }
+
+    /**
+     * FIPS mode will be enabled only if the semeru.fips system
+     * property is true (default as false).
+     *
+     * @return true if FIPS is enabled
+     */
+    public static boolean enableFIPS() {
+        return shouldEnableFIPS;
+    }
+
+    /**
+     * Remove the security providers and only add the FIPS security providers.
+     *
+     * @param props the java.security properties
+     * @return true if the FIPS properties loaded successfully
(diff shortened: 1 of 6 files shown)


WilburZjh:
Fix issues in #590

WilburZjh:
@keithc-ca @pshipton @alon-sh

jdk/src/share/classes/java/security/Security.java
```diff
@@ -187,6 +195,18 @@ private static void initialize() {
...
+            boolean fipsEnabled = FIPSConfigurator.configureFIPS(props);
+            if (sdebug != null) {
+                if (fipsEnabled) {
+                    sdebug.println("FIPS mode enabled.");
+                } else {
+                    sdebug.println("FIPS mode disabled.");
+                }
+            }
+        }
+
```
keithc-ca, line 209: Please remove this blank line (that wasn't in the earlier pull request).

closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,159 @@
...
+                        }
+                        i.remove();
+                    }
+                }
+
+                // Add FIPS security providers.
+                props.put("security.provider.1", "sun.security.pkcs11.SunPKCS11 ${java.home}/lib/security/nss.fips.cfg");
+                props.put("security.provider.2", "sun.security.provider.Sun");
+                props.put("security.provider.3", "sun.security.ec.SunEC");
+                props.put("security.provider.4", "com.sun.net.ssl.internal.ssl.Provider");
```
keithc-ca, line 113: I notice the values here appear to be fully-qualified class names; this is different than the previous change and different than other jdk versions. Are fully-qualified class names also needed in new versions? If not, why not?

keithc-ca (changes requested):

jdk/src/share/classes/java/security/Security.java
WilburZjh: removed

closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
WilburZjh: In JDK8,  for SunPKCS11 provider, in the [ProviderConfig.java](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/sun/security/jca/ProviderConfig.java) file at line [81](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/sun/security/jca/ProviderConfig.java#L81), it checks if the provider name equals to line [49](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/sun/security/jca/ProviderConfig.java#L49). Therefore, in JDK8, we need a full package name for provider `SunPKCS11`. In JDK17, for SunPKCS11 provider, in the [ProviderConfig.java](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/java.base/share/classes/sun/security/jca/ProviderConfig.java) file at line [76](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/java.base/share/classes/sun/security/jca/ProviderConfig.java#L76), it checks if the provider name ends with `SunPKCS11`. Therefore, for SunPKCS11, we need to full package name in JDK8.

`SunJCE` is an abstract class in JDK8, [Provider](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/com/sun/net/ssl/internal/ssl/Provider.java) should be used as SunJCE. In JDK17, there is an class for [SunJCE](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/55387a2d979e3fe83e8fbb3f900f013bf4775cbc/src/java.base/share/classes/com/sun/crypto/provider/SunJCE.java).

All the providers are defined with full package names in the `java.security` file for JDK8. 
```
security.provider.1=sun.security.provider.Sun
security.provider.2=sun.security.rsa.SunRsaSign
security.provider.3=sun.security.ec.SunEC
security.provider.4=com.sun.net.ssl.internal.ssl.Provider
security.provider.5=com.sun.crypto.provider.SunJCE
security.provider.6=sun.security.jgss.SunProvider
security.provider.7=com.sun.security.sasl.Provider
security.provider.8=org.jcp.xml.dsig.internal.dom.XMLDSigRI
security.provider.9=sun.security.smartcardio.SunPCSC
```
In contrast, all the providers are defined with suffix in the `java.security` file for JDK17.
```
security.provider.1=SUN
security.provider.2=SunRsaSign
security.provider.3=SunEC
security.provider.4=SunJSSE
security.provider.5=SunJCE
security.provider.6=SunJGSS
security.provider.7=SunSASL
security.provider.8=XMLDSig
security.provider.9=SunPCSC
security.provider.10=JdkLDAP
security.provider.11=JdkSASL
security.provider.12=SunPKCS11
```

pshipton (approved):

closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
keithc-ca: Looking further into this difference, it seems jdk8 needs fully-qualified class names (see [ProviderConfig.doLoadProvider()](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/sun/security/jca/ProviderConfig.java#L214-L218)) while jdk17 accepts either simple names (e.g. "SUN") or fully-qualified names (e.g. "sun.security.provider.Sun"), see [ProviderConfig.doLoadProvider()](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/java.base/share/classes/sun/security/jca/ProviderConfig.java#L181). In jdk17, the comments in [java.security](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/java.base/share/conf/security/java.security#L34-L54) seems to suggest that simple provider names are preferred to fully-qualified class names.

keithc-ca:
jenkins test sanity xlinux jdk8

closed/adds/jdk/src/share/classes/openj9/internal/security/FIPSConfigurator.java
pshipton: Be that as it may, the java.security file provided for jdk11+ uses the short names.
```
security.provider.1=SUN
security.provider.2=SunRsaSign
security.provider.3=SunEC
security.provider.4=SunJSSE
...
```

keithc-ca (approved):

PR #602, opened 2022-08-16
Fix incremental DDR builds

keithc-ca:
`ClassGenerator` reads `CompatibilityConstants29.dat` and so must be executed if that file changes.
The preprocessor must also be re-run to propagate any changes.


pshipton:
jenkins compile all jdk8

pshipton (approved):

PR #615, opened 2022-09-27
Do not use short name on Windows for vcruntime140.dll

pshipton:
This change is also seen in jdk11
https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/be05050

Testing ongoing in https://openj9-jenkins.osuosl.org/job/Pipeline-Build-Test-Personal/250/
It's not done, but the configure output looks good.

Issue https://github.com/eclipse-openj9/openj9/issues/15962


keithc-ca:
This is essentially the same as #616 (I didn't realize you were working on this), minus the copyright notice.

keithc-ca:
> configure output looks good

And the build output shows the right name is included in images
```
15:55:15  j2sdk-image/bin/vcruntime140.dll
15:55:17  j2sdk-image/jre/bin/vcruntime140.dll
```
Can you fix the missing copyright?

pshipton:
> Can you fix the missing copyright?

Since the change can be considered backported from jdk11, I didn't add an IBM copyright.

pshipton:
btw, I don't think this fixes https://github.com/eclipse-openj9/openj9/issues/15962, but it's part 1 to get it working. There is also this part:
"
After going through preloadLibrary code, we figured it might be a problem with the way the preload libraries are being Loaded.

https://github.com/eclipse-openj9/openj9/blob/v0.27.0-release/runtime/j9vm/jvm.c#:~:text=LOAD_WITH_ALTERED_SEARCH_PATH

Since, altered searched strategy is used instead of standard search strategy, vcruntime140.dll is being searched in dll directory (JRE_HOME/bin/default/) instead of application directory (JRE_HOME/bin/) for certain LoadLibraryEx calls.
"

keithc-ca (approved):

PR #616, opened 2022-09-27
Fix name of vcruntime140.dll

keithc-ca:
Don't "fixup" path to dll; backported from
* 8201267: Disable warnings for VS2017 to enable building

Fixes: eclipse-openj9/openj9#15962.


keithc-ca:
This is a duplicate of #615.

Issue #611, opened 2022-09-21
Update the OpenJDK exclude tests list for FIPS


WilburZjh:
The FIPS mode has restrictive security configurations, so update the OpenJDK exclude tests list “ProblemList-fips.txt” for the FIPS testing.

WilburZjh:
### Exclude tests list from sanity.openjdk

**java.security.NoSuchProviderException: no such provider: SunRsaSign** 

- java/math/BigInteger/ModPow65537.java.ModPow65537

**Existing Jars sign related**

- java/util/jar/JarFile/ScanSignedJar.java.ScanSignedJar
- java/util/jar/JarFile/TurkCert.java.TurkCert
- java/util/jar/JarInputStream/ScanSignedJar.java.ScanSignedJar
- java/util/jar/JarInputStream/TestIndexedJarWithBadSignature.java.TestIndexedJarWithBadSignature

**-keystore ks -storepass changeit -keypass changeit -keyalg rsa -alias a -dname CN=A -genkeypair
keytool error: java.security.KeyStoreException: 
sun.security.pkcs11.wrapper.PKCS11Exception: CKR_ATTRIBUTE_VALUE_INVALID**

- java/util/jar/JarInputStream/ExtraFileInMetaInf.java.ExtraFileInMetaInf

WilburZjh:
### Exclude tests list from extended.openjdk

**java.security.KeyStoreException: JCEKS not found**

- sun/security/tools/jarsigner/TsacertOptionTest.java.TsacertOptionTest

**java.security.KeyStoreException: JKS not found
java.security.NoSuchAlgorithmException: JKS KeyStore not available**

- java/security/KeyStore/CheckInputStream.java.CheckInputStream
- java/security/KeyStore/KeyStoreBuilder.java.KeyStoreBuilder
- java/security/KeyStore/PBETest.java.PBETest
- java/security/KeyStore/PKCS12/CheckDefaults.java.CheckDefaults
- java/security/KeyStore/PKCS12/WriteP12Test.java.WriteP12Test
- java/security/KeyStore/TestKeyStoreBasic.java.TestKeyStoreBasic
- java/security/KeyStore/TestKeystoreCompat.java.TestKeystoreCompat
- java/security/Policy/SignedJar/SignedJarTest.java.SignedJarTest
- javax/net/ssl/ALPN/SSLEngineAlpnTest.java.SSLEngineAlpnTest
[...]

PR #619, opened 2022-09-29
Update excluded list for extended.openjdk

WilburZjh:
Signed-off-by: Jinhang Zhang <Jinhang.Zhang@ibm.com>

### DIFF
--- a/jdk/test/ProblemList-fips.txt
+++ b/jdk/test/ProblemList-fips.txt
@@ -859,3 +859,20 @@ sun/security/pkcs11/Mac/MacSameTest.java                            https://gith
 sun/security/pkcs11/Mac/ReinitMac.java                              https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611   linux-x64
 sun/security/pkcs11/tls/TestPRF.java                                https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611   linux-x64
 sun/security/ssl/internal/TestRun.sh                                https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611   linux-x64
+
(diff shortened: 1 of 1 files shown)


WilburZjh:
Updated two failed tests

- [javax/crypto/CipherSpi/TestGCMWithByteBuffer.java](https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611#issuecomment-1262741995)
- [sun/rmi/rmic/manifestClassPath/run.sh](https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611#issuecomment-1262791680)

pshipton:
@alon-sh for review

llxia:
@alon-sh could you please review the PR when you have time? Thanks

alon-sh:
the first test is fine, the second test that is excluded:

```
sun/rmi/rmic/manifestClassPath/run.sh                               https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611   linux-x64
```

@WilburZjh could you give more details on why is this test failing and why should it be excluded rather then fixed?

WilburZjh:
> the first test is fine, the second test that is excluded:
> 
> ```
> sun/rmi/rmic/manifestClassPath/run.sh                               https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/611   linux-x64
> ```
> 
> @WilburZjh could you give more details on why is this test failing and why should it be excluded rather then fixed?

This jvm option has been removed from other jdk versions. So Im not sure if we need to remove it as well or fix it.

alon-sh:
ok you can keep it excluded. approved.

fyi @llxia @pshipton

pshipton (approved):

PR #634, opened 2022-12-06
Allow building on macOS with XCode 14

kgibm:


kgibm:
@pshipton

pshipton:
@keithc-ca should we be allowing the change in permissions for autogen.sh?

keithc-ca:
> @keithc-ca should we be allowing the change in permissions for autogen.sh?

No, I don't think we want that change.

pshipton:
> No, I don't think we want that change.

That's what I figured, what I'm not sure about is why not?

kgibm:
I just made that change for convenience. Alternatively, I could have executed the script using `sh`. I'm okay with reverting it.

Another question I had around the copyright check failure. I noticed that some of these files have multiple IBM copyrights. Is that correct? For example:

```
$ grep "Copyright.*IBM" common/autoconf/generated-configure.sh
# (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved
# (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
# (c) Copyright IBM Corp. 2020, 2021 All Rights Reserved
# (c) Copyright IBM Corp. 2021, 2022 All Rights Reserved
# (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
```

keithc-ca:
> > No, I don't think we want that change.
> 
> That's what I figured, what I'm not sure about is why not?

I recall someone else proposing a similar change (in another jdk version). The documented use-case didn't require the file have execute permission - I suggested that changing the permission was something to be proposed upstream. I couldn't find documentation that applies here, but I think the same response is appropriate.

keithc-ca:
> some of these files have multiple IBM copyrights. Is that correct?

Yes, that's expected (and not flagged by the copyright check job). That file in particular is generated from multiple source files, many of which have an IBM copyright notice.

kgibm:
Nevermind, I realize that reverting the permission change also avoids the copyright check on that file.

kgibm:
Yes, successfully built yesterday:

```
$ pkgutil --pkg-info=com.apple.pkg.CLTools_Executables | grep version
version: 14.0.0.0.1.1661618636
$ ( cd build/macosx-x86_64-normal-server-release/images/j2sdk-image; ./bin/java -version)
openjdk version "1.8.0_362-internal"
OpenJDK Runtime Environment (build 1.8.0_362-internal-kevin_2022_12_06_13_39-b00)
Eclipse OpenJ9 VM (build master-504fd6a3e, JRE 1.8.0 Mac OS X amd64-64-Bit Compressed References 20221206_000000 (JIT enabled, AOT enabled)
OpenJ9   - 504fd6a3e
OMR      - 90eea450e
JCL      - a441a5a019 based on jdk8u362-b05)
```

And then I realized that I needed to do Linux builds for testing the OMR PR, not macOS 🤦

keithc-ca (approved):

keithc-ca:
Jenkins compile osx jdk8

PR #635, opened 2022-12-12
update exclude list for fips

WilburZjh:
Closes https://github.com/eclipse-openj9/openj9/issues/16412


pshipton:
@WilburZjh your commit is not signed and is failing the check. Pls update the commit with a Signed-off-by.

WilburZjh:
updated

pshipton:
It doesn't look like there is a blank line between the title and the Signed-off-by line. There should be a blank line before the Signed-off-by.

pshipton:
Ok, I see there is a blank line, it's just not obvious when looking via github.

PR #639, opened 2022-12-14
Fail test-image-openj9 target when java -version fails

jdmpapin:
Previously, the exit status would come from `tee`, which succeeded.

### DIFF
--- a/closed/make/Main.gmk
+++ b/closed/make/Main.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2023 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
@@ -63,7 +63,7 @@ test-image : test-image-openj9
 test-image-openj9 : images
 	@+$(MAKE) $(MAKE_ARGS) -f $(SRC_ROOT)/closed/TestImage.gmk
 ifneq ($(COMPILE_TYPE), cross)
-	$(JRE_IMAGE_DIR)/bin/java -version 2>&1 | $(TEE) $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt
+	rc=0; $(JRE_IMAGE_DIR)/bin/java -version >$(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt 2>&1 || rc="$$?"; $(CAT) $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt; exit "$$rc"
 endif
 
 clean-docs : clean-openj9-only-docs


jdmpapin:
Does this repo still need to require commit sign-off for some reason? I suspect that it doesn't, and that the failing check is a holdover

keithc-ca:
If you have good a reason that the check should be removed, then please capture it in an issue. Until we have agreement to remove the signed-off-by check, please update the commit message.

closed/make/Main.gmk
```diff
@@ -63,7 +63,7 @@ test-image : test-image-openj9
 test-image-openj9 : images
 	@+$(MAKE) $(MAKE_ARGS) -f $(SRC_ROOT)/closed/TestImage.gmk
 ifneq ($(COMPILE_TYPE), cross)
-	$(JRE_IMAGE_DIR)/bin/java -version 2>&1 | $(TEE) $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt
+	$(JRE_IMAGE_DIR)/bin/java -version >$(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt 2>&1; rc=$$?; cat $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt; exit "$$rc"
```
keithc-ca, line 66: Perhaps setting `pipefail` would be a better way to address this.

jdmpapin: Do we know that `pipefail` is always available? I wasn't sure about that, which is why I wrote it this way. If someone can vouch though that we have `pipefail` on all platforms (i.e. we won't be running a plain POSIX `sh`, etc.), then I'd be happy to use that instead.

jdmpapin:
> Until we have agreement to remove the signed-off-by check, please update the commit message.

Updated

> If you have good a reason that the check should be removed, then please capture it in an issue.

Opened #644

closed/make/Main.gmk
jdmpapin: The version is printed to stderr, so redirecting to the file in a second run means redirecting stderr. With stderr redirected, any output indicating a crash in that second run would be hidden from the build log

keithc-ca: I don't understand your concern. When I said "run it twice", I meant something like
```
java -version 2>&1
java -version >version.txt 2>&1
```
If the first fails, we'll see the output. If the first succeeds, the second is likely to as well and the output should be the same. Yes, if the second fails we won't see its output, but I expect that to be an exceedingly rare situation.

jdmpapin: > Yes, if the second fails we won't see its output

This was my concern.

> but I expect that to be an exceedingly rare situation.

I do too, but that almost makes it worse. I think it's likely enough to happen at some point, and it will be extremely confusing

jdmpapin: I don't think we should make failures confusing just because they're rare.

keithc-ca: The second run could continue to pipe its output through `$(TEE)`. We would get two copies of the output, but I could live with that.

jdmpapin: That would be the same as the current command, i.e. it could crash without causing `make` to fail

keithc-ca: Ok, to be robust, perhaps we need to take your approach. Please replace `cat` with `$(CAT)`.

jdmpapin:
Changed `cat` to `$(CAT)` as requested.

keithc-ca (approved):

keithc-ca:
Do you intend to offer equivalent changes for other jdk versions (11, 17, 19, 20, next)?

jdmpapin:
Oh, thanks, good catch! I should, yes. When I [built various versions on my local machine under `strace`](https://github.com/eclipse-openj9/openj9/pull/16426#issuecomment-1341735017), I only noticed this `java -version` execution in Java 8, since there it was the only execution of a Java binary from within build/. I thought that it didn't happen in later versions, but it seems that it does and I just didn't notice it. I can see the rule is still there in Main.gmk, and I just re-ran Java 17 `make` under `strace` to see that `java -version` does run.

jdmpapin:
There a 'next' extensions repo? I don't see one.

jdmpapin:
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/614
https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/177
https://github.com/ibmruntimes/openj9-openjdk-jdk19/pull/62
https://github.com/ibmruntimes/openj9-openjdk-jdk20/pull/10
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/543

jdmpapin:
Updated this PR so that the changes are consistent with those for later JDK versions

closed/make/Main.gmk
```diff
@@ -63,7 +63,7 @@ test-image : test-image-openj9
 test-image-openj9 : images
 	@+$(MAKE) $(MAKE_ARGS) -f $(SRC_ROOT)/closed/TestImage.gmk
 ifneq ($(COMPILE_TYPE), cross)
-	$(JRE_IMAGE_DIR)/bin/java -version 2>&1 | $(TEE) $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt
+	rc=0; $(JRE_IMAGE_DIR)/bin/java -version >$(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt 2>&1 || rc=$$?; $(CAT) $(IMAGES_OUTPUTDIR)/test/openj9/java-version.txt; exit "$$rc"
```
keithc-ca, line 66: Sorry to pick nits, but why the quotes in `exit "$$rc"` (if they're not needed in `rc=$$?`, they're not needed here either)?

jdmpapin: That's a habit of mine. I think the way I'd put it is that if you see `$?`, quotes are never necessary, and you don't need to read any other part of the script to figure that out. For `$rc`, we can reason that they quotes are unnecessary, but we have to do that reasoning, and with later changes that reasoning could conceivably get out of date (though it's admittedly unlikely). So I always quote my own variables.

If for some reason these two variable uses absolutely must be consistent (i.e. both quoted or both unquoted), then I'll add quotes around `$?`. But if possible I'd prefer not to have to make that change. I don't think it's very helpful, and it's extremely tedious to make another trivial change in these 6 PRs and then check all of them to make sure I'm still getting the behaviour I expect.

keithc-ca: `$?` should expand to a number and `exit` expects a single, numeric argument, so there's not needed in either place: My preference is that we don't use quotes here. If you want to use quotes, please do so consistently (i.e. both places).

jdmpapin: Fine

jdmpapin:
Added quotes for consistency

closed/make/Main.gmk
keithc-ca: Perhaps I should have insisted on removing the quotes; they're only used in places where they're provably unnecessary (and it's still inconsistent: the other assignment to `rc` doesn't have them). I don't think it's worth another round of changes, so let's declare victory.

keithc-ca:
Jenkins compile zlinux jdk8

closed/make/Main.gmk
jdmpapin: Expanding variables outside of double-quoted strings as much as possible is an error-prone way to write shell scripts. The only reliable way to consistently treat uses of variables (w.r.t. quoting) is to always quote them. That's why that's my general rule (though I sometimes make an exception for `$?` only because it's obvious that it's OK even without reading any other part of the script).

And I disagree that my updated code is still inconsistent in this respect. The literal string `0` is unquoted, whereas uses of variables are quoted. It's irrelevant that both a literal and a variable occur in assignments.

I can at least agree with this though:
> I don't think it's worth another round of changes, so let's declare victory.

keithc-ca (approved):

PR #649, opened 2023-02-14
OpenJ9 doesn't support Perf, don't try to use it

pshipton:
Trying to use sun.misc.Perf may result in an Assertion "Java_sun_misc_Perf_attach unimplemented" or similar. Calling sun.tools.jconsole.LocalVirtualMachine.getAllVirtualMachines() tries to use Perf.

Issue https://github.com/eclipse-openj9/openj9/issues/16719


keithc-ca:
Jenkins compile alinux64 jdk8

keithc-ca (approved):

PR #652, opened 2023-03-13
Expand OpenSSL Version granularity

WilburZjh:
Additional granularity for tracking versions is needed in Native code to make better branching decisions based upon the library versions. Currently, there is differentiation only between <=1.0.x and >=1.1.0. Version 1.1.1 will be supported until 2023-09-11,
versions including 1.1.0, 1.0.2, 1.0.1, 1.0.0, and 0.9.8 are no longer supported. We need to add support for separate versions to deprecate/add features in the future.

### DIFF
--- a/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


WilburZjh:
@keithc-ca @pshipton please help to review

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -50,18 +50,17 @@ public class NativeCrypto {
     private static final boolean traceEnabled = Boolean.parseBoolean(
             GetPropertyAction.privilegedGetProperty("jdk.nativeCryptoTrace", "false"));
 
-    //ossl_ver:
+    //ossl_vers will be either:
     // -1 : library load failed
-    //  0 : openssl 1.0.x
-    //  1 : openssl 1.1.x or newer
+    // or one of the OPENSSL_VERSION_x_x_x constants
```
keithc-ca, line 55: Why are those "OPENSSL_VERSION_x_x_x constants" not declared here?
Please add them or update the comment to make it clear those constants are declared in `NativeCrypto.c`.

keithc-ca (changes requested):

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
WilburZjh: Added.

keithc-ca:

PR #665, opened 2023-05-08
Remove useless patterns from .gitignore

keithc-ca:
* configure option `--with-freetype-src` has been removed, so `freetype.bat` and `freetype.log` should not be created

See also https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/654.


JasonFengJ9 (approved):
LGTM

No PR build required

PR #670, opened 2023-05-25
Don't expect copyright notices in policy or security files

keithc-ca:
Fixes false positive in #663.

Also tidy up.

### DIFF
--- a/.copyrightignore
+++ b/.copyrightignore
@@ -1,12 +1,12 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
 # ===========================================================================
 #
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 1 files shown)


keithc-ca:
Ping @AdamBrousseau

.copyrightignore
```diff
@@ -20,31 +20,34 @@
...
+# 6. *.ignore # all ignore files only
+# 7. **/jobs/**/pipelines/ # directory path contains both jobs and pipelines not root directory
+# 8. *job* # all things at contains characters in between stars
 # ===========================================================================
 
 src/bsd/doc
 src/jdk.crypto.ec/share/native/libsunec/impl
 src/linux/doc
+src/share/lib/security/java.policy
+src/share/lib/security/java.security-*
```
AdamBrousseau, line 38: Do we know why these aren't prepended with `jdk/` part of the path?

keithc-ca: I followed the existing pattern: paths like this should be ignored whether they're within `jdk` or within `closed/overrides/jdk`.

AdamBrousseau (approved):

PR #677, opened 2023-06-30
Improve speed of cipher operations for large inputs and allow unpadding at offset of output

KostasTsiounis:
In cases were bytes remain buffered and/or padding is enabled and encryption is performed, the entirety of the input provided in the `doFinal()` methods of the `CipherCore` class would be copied in a buffer with enough space to accommodate the extra bytes requires for buffered bytes and/or padding. The buffer would subsequently be zeroed out for security reasons.

To avoid said actions and improve the performance of the process in case of very large inputs, the updated code calls the `update()` method with as much data as possible before the final steps.

Changes made as part of https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/d8fc6df82151f89566508fcb9da1e18f9a236527 had to, also, be back-ported to allow unpadding to happen at a specific offset of the output. Those involved adding an extra parameter in the `unpad()` method and using it when invoking the corresponding method of the assigned Padding.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/614

### DIFF
--- a/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/jdk/src/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins test sanity.openjdk zlinux jdk8

keithc-ca (approved):

PR #678, opened 2023-07-04
macOS: Stop using sprintf in java.desktop

knn-k:
This commit backports the changes for macOS from Java 21, replacing sprintf() by snprintf().

### DIFF
--- a/jdk/src/macosx/native/com/sun/media/sound/PLATFORM_API_MacOSX_Ports.cpp
+++ b/jdk/src/macosx/native/com/sun/media/sound/PLATFORM_API_MacOSX_Ports.cpp
@@ -619,7 +619,7 @@ void PORT_GetControls(void* id, INT32 portIndex, PortControlCreator* creator) {
                     CFRelease(cfname);
                 } else {
                     channelName = (char *)malloc(16);
-                    sprintf(channelName, "Ch %d", ch);
+                    snprintf(channelName, 16, "Ch %d", ch);
                 }
 
                 void* jControls[2];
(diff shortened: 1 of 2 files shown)


jdk/src/macosx/native/com/sun/media/sound/PLATFORM_API_MacOSX_Ports.cpp
```diff
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (C) Copyright IBM Corp. 2023 All Rights Reserved.
```
pshipton, line 27: Since the change is an OpenJDK backport, we shouldn't be adding an IBM copyright.

knn-k: I removed the IBM copyright.

jdk/src/macosx/native/sun/font/AWTStrike.m
```diff
@@ -107,7 +107,7 @@ + (AWTStrike *) awtStrikeForFont:(AWTFont *)awtFont
 #define AWT_FONT_CLEANUP_FINISH                                         \
     if (_fontThrowJavaException == YES) {                               \
         char s[512];                                                    \
-        sprintf(s, "%s-%s:%d", THIS_FILE, __FUNCTION__, __LINE__);       \
+        snprintf(s, sizeof(s), "%s-%s:%d", __FILE__, __FUNCTION__, __LINE__);       \
```
pshipton, line 110: For consistency, I don't think the THIS_FILE change should be made. It was removed on the way to jdk11 by OpenJDK 8226346 (https:/github.com/ibmruntimes/openj9-openjdk-jdk11/commit/8fab614110), this OpenJDK issue isn't public. When THIS_FILE was removed, it was done as part of a bigger change to a number of files and not just this one. 

https:/github.com/ibmruntimes/openj9-openjdk-jdk11/commit/8fab614110

knn-k: Fixed as suggested.

pshipton (approved):

PR #680, opened 2023-07-19
Merge jdk8u382-b05 into 0.40.0 release

keithc-ca:
~Draft until the expected jdk8u382-ga tag appears.~ Let's assume the ga tag will line up with the b05 tag.


JasonFengJ9 (approved):
> Let's assume the ga tag will line up with the b05 tag.

If not, the newer tag can be pulled into `0.40.0` branch afterwards.

pshipton:
The ga tag is created at the b05 level.
https://github.com/openjdk/jdk8u/releases/tag/jdk8u382-ga

PR #692, opened 2023-09-12
Use the system zlib on Linux platform

pshipton:
This makes jdk8 consistent with jdk11+.

Issue https://github.com/eclipse-openj9/openj9/issues/18127


pshipton:
jenkins compile aix,zlinux jdk8

pshipton:
AIX `checking for which zlib to use... system`
zlinux `checking for which zlib to use... system`

pshipton:
The PR build is https://openj9-jenkins.osuosl.org/job/PullRequest-OpenJDK8/93/
Pushing a change to the commit wording.

common/autoconf/libraries.m4
```diff
@@ -668,9 +668,9 @@ AC_DEFUN_ONCE([LIB_SETUP_MISC_LIBS],
   AC_MSG_CHECKING([for which zlib to use])
 
   DEFAULT_ZLIB=bundled
-  if test "x$OPENJDK_TARGET_OS" = xmacosx -o "x$OPENJDK_TARGET_OS" = xaix; then
+  if test "x$OPENJDK_TARGET_OS" = xmacosx -o "x$OPENJDK_TARGET_OS" = xaix -o "x$OPENJDK_TARGET_OS" = xlinux; then
     #
-    # On macosx and aix default is system...on others default is
+    # On macosx, aix, and linux default is system...on others default is
```
keithc-ca, line 673: Since you're changing this line, could you please complete it?
```
    # On macosx, aix, and linux default is system...on others default is bundled.
```

pshipton: Updated.

common/autoconf/libraries.m4
```diff
@@ -668,9 +668,9 @@ AC_DEFUN_ONCE([LIB_SETUP_MISC_LIBS],
   AC_MSG_CHECKING([for which zlib to use])
 
   DEFAULT_ZLIB=bundled
-  if test "x$OPENJDK_TARGET_OS" = xmacosx -o "x$OPENJDK_TARGET_OS" = xaix; then
+  if test "x$OPENJDK_TARGET_OS" = xmacosx -o "x$OPENJDK_TARGET_OS" = xaix -o "x$OPENJDK_TARGET_OS" = xlinux; then
     #
-    # On macosx and aix default is system...on others default is
+    # On macosx, aix, and linux default is system...on others default is bundled.
```
keithc-ca, line 673: You say this "makes jdk8 consistent with jdk11+", but the consistency is not as clear as it could be. In other versions, the starting position is `system` with exceptions for Windows (11+) and macOS aarch64 (17+) where it is `bundled`.

Should we make jdk11 consistent with later versions (by bundling on macOS aarch64)?

pshipton: There is no amac jdk8, but I can change the default to system with an exception for Windows if you prefer. I must have made a mistake as my chart shows amac as system for jdk17, but I checked builds and see it's bundled as you say.

pshipton: Actually, something changed. The 0.40 build has amac jdk17 as system, but current openj9 builds and the 0.41 M1 build used bundled.

pshipton: I'm also less concerned with Mac, I could change the wording to be Linux specific "This makes jdk8 Linux consistent with jdk11+".

keithc-ca: The change was perhaps due to https://github.com/adoptium/temurin-build/pull/3459 which has since been reverted for Semeru builds (see https://github.com/ibmruntimes/temurin-build/pull/88).

I was suggesting that here, the test should mirror that of later versions (without the irrelevant consideration for macOS aarch64), like `lib-bundled.m4` says in jdk11:
```
  DEFAULT_ZLIB=system
  if test "x$OPENJDK_TARGET_OS" = xwindows; then
    # On windows default is bundled...on others default is system
    DEFAULT_ZLIB=bundled
  fi
```
but that jdk11 should be updated with the check for macOS aarch64 like jdk17+.

pshipton: Updated. Will modify jdk11 as well.

pshipton: Created https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/705

keithc-ca (approved):

keithc-ca:
In https://openj9-jenkins.osuosl.org/job/Build_JDK8_x86-32_windows_Personal/737/consoleText:
```
[2023-09-13T14:43:21.084Z] checking for which zlib to use... bundled
```
In https://openj9-jenkins.osuosl.org/job/Build_JDK8_x86-64_linux_Personal/462/consoleText:
```
[2023-09-13T14:42:44.156Z] checking for which zlib to use... system
```

PR #714, opened 2023-11-30
Add Redhat NSS FIPS support on p/z linux platforms

taoliult:
The Redhat NSS FIPS module is certified on ppc64le and s390x linux platforms. Adding the Redhat NSS FIPS support for the Semeru OpenJDK on these two platforms.

This is a back-port PR from JDKNext PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/703


taoliult:
@pshipton Please help to review.

pshipton:
This doesn't have the java.security changes, is that expected?

taoliult:
@pshipton 

Yes, I think it is expected. 

Because in OpenJDK8, there are the "java.security" files for different platforms. For example, jdk/src/share/lib/security/java.security-linux, 
jdk/src/share/lib/security/java.security-aix, 
jdk/src/share/lib/security/java.security-macosx, and etc. 

And also, there is no "MakeJavaSecurity.java" to define the "#ifdef, #ifndef and #endif". So, I don't think we need to make changes for "java.security" file.

pshipton:
jenkins compile zlinux jdk8

pshipton:
I see. The RestrictedSecurity settings are already there on Linux platforms.

pshipton (approved):

PR #753, opened 2024-05-13
Throw OOM if array size exceeds int maximum

theresa-m:
Java 11: https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/782
Java 17: https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/352
Java 21: https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/158
Java 22: https://github.com/ibmruntimes/openj9-openjdk-jdk22/pull/59
Java next: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/791

### DIFF
--- a/jdk/src/share/classes/java/lang/StringCoding.java
+++ b/jdk/src/share/classes/java/lang/StringCoding.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.lang;
 
 import java.io.UnsupportedEncodingException;
@@ -92,7 +98,11 @@ class StringCoding {
     private static int scale(int len, float expansionFactor) {
         // We need to perform double, not float, arithmetic; otherwise
         // we lose low order bits when len is larger than 2**24.
-        return (int)(len * (double)expansionFactor);
+        double result = len * (double)expansionFactor;
+        if (result > (double)Integer.MAX_VALUE) {
+            throw new OutOfMemoryError("Requested array size exceeds limit");
+        }
+        return (int)result;
     }
 
     private static Charset lookupCharset(String csn) {


theresa-m:
@pshipton can you please review?

pshipton:
If I understand correctly, before this change a different exception would be thrown, an ArrayIndexOutOfBounds or NegativeArrayIndexOutOfBounds. Throwing OutOfMemoryError clarifies the situation.

jdk/src/share/classes/java/lang/StringCoding.java
```diff
@@ -92,7 +98,11 @@ private static char[] safeTrim(char[] ca, int len,
     private static int scale(int len, float expansionFactor) {
         // We need to perform double, not float, arithmetic; otherwise
         // we lose low order bits when len is larger than 2**24.
-        return (int)(len * (double)expansionFactor);
+        double result = len * (double) expansionFactor;
```
keithc-ca, line 101: Please don't change the formatting of the cast:
```java
        double result = len * (double)expansionFactor;
```

keithc-ca (changes requested):

keithc-ca:
This problem is not unique to Java 8; see the [head stream](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/src/java.base/share/classes/java/lang/String.java#L1061).

theresa-m:
> If I understand correctly, before this change a different exception would be thrown, an ArrayIndexOutOfBounds or NegativeArrayIndexOutOfBounds. Throwing OutOfMemoryError clarifies the situation.

Yes thats correct, eventually [the array index dp will overflow](https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/jdk/src/share/classes/sun/nio/cs/UTF_8.java#L705) since the method expects the string will fit in the allocated array. This change provides a more user friendly error that better describes the problem.

jdk/src/share/classes/java/lang/StringCoding.java
theresa-m: Thanks Keith, I removed the added space.

theresa-m:
@keithc-ca I have linked ports to the 11, 17, 21, 22 and next extensions repo in the summary. Can you please review?

keithc-ca (approved):

PR #755, opened 2024-05-28
Get native EC key pointer during init through ECUtil instead of key impl

KostasTsiounis:
The acquisition of a native pointer for an EC key, private or public, is being moved to the ECUtil, instead of the implementation classes of said keys. This allows use of interop keys, rather than just keys coming from SunEC.

The methods to get the pointers to the native keys are called during the initialization phases, to discover the possibility of not being able to utilize the native code and revert to the Java implementation early.

The engineUpdate() methods of the RawECDSA subclass are, also, updated to use the java implementation, if one has been initialized.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/792

Signed-off by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

### DIFF
--- a/jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java
+++ b/jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java
@@ -23,12 +23,6 @@
  * questions.
  */
 
-/*
- * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
- * ===========================================================================
- */
-
 package sun.security.ec;
 
 import java.io.IOException;
@@ -39,8 +33,6 @@ import java.math.BigInteger;
 import java.security.*;
 import java.security.interfaces.*;
 import java.security.spec.*;
-
-import jdk.crypto.jniprovider.NativeCrypto;
 
 import sun.security.util.ArrayUtil;
 import sun.security.util.DerInputStream;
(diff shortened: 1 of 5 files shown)


KostasTsiounis:
FYI @keithc-ca

jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java
```diff
@@ -240,36 +230,4 @@ private void readObject(ObjectInputStream stream)
...
-                        }
-                    }
-                    this.nativeECKey = nativePointer;
-                }
-            }
-        }
-        return this.nativeECKey;
-    }
-}
+}
```
keithc-ca, line 233: Please restore this newline.

jdk/src/share/classes/sun/security/ec/ECPublicKeyImpl.java
```diff
@@ -156,41 +145,4 @@ private void readObject(ObjectInputStream stream)
...
-                        }
-                    }
-                    this.nativeECKey = nativePointer;
-                }
-            }
-        }
-        return this.nativeECKey;
-    }
-}
+}
```
keithc-ca, line 148: Please restore this newline.

jdk/src/share/classes/sun/security/ec/ECPrivateKeyImpl.java
KostasTsiounis: Done.

jdk/src/share/classes/sun/security/ec/ECPublicKeyImpl.java
KostasTsiounis: Done.

keithc-ca:
Jenkins test sanity win32 jdk8

keithc-ca (approved):

PR #756, opened 2024-06-20
Allow extension of RestrictedSecurity profiles

KostasTsiounis:
`RestrictedSecurity` profiles sometimes share a lot of duplicate settings with only minor differences. With these changes the extension, similar to object-orientation, of profiles becomes possible.

More specifically, a profile `A` can extend another a profile `B`, using `RestrictedSecurity.<profile A name>.extends = RestrictedSecurity.<profile B name>`. This allows profile `A` to inherit all of profile `B`'s properties. One can add additional properties to profile `A`, or amend some of the existing ones. That includes overriding, appending or removing from a property (wherever that's applicable).

An additional property is introduced. The `RestrictedSecurity.<profile name>.desc.hash = <hash algorithm>:<hash>` is used to ensure the profile hasn't been unintentionally altered. The profile's properties are hashed using the selected `<hash algorithm>`, and the result is compared to the `<hash>` provided through the property. This property is mandatory for base profiles (i.e., profiles that are not extending anything), and optional for the rest.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/793

### DIFF
--- a/closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -23,7 +23,10 @@
  */
 package openj9.internal.security;
 
+import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
 import java.security.Provider.Service;
 import java.time.LocalDate;
@@ -43,6 +46,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import sun.security.util.Debug;
 
@@ -71,6 +76,8 @@ public final class RestrictedSecurity {
     private static boolean securityEnabled;
 
     private static String userSecurityID;
+
+    private static ProfileParser profileParser;
 
     private static RestrictedSecurityProperties restricts;
 
@@ -163,6 +170,20 @@ public final class RestrictedSecurity {
 
     private RestrictedSecurity() {
         super();
+    }
+
+    /**
+     * Check loaded profiles' hash values.
+     *
+     * In order to avoid unintentional changes in profiles and incentivize
+     * extending profiles, instead of altering them, a digest of the profile
+     * is calculated and compared to the expected value.
+     */
+    public static void checkHashValues() {
+        if (profileParser != null) {
+            profileParser.checkHashValues();
+            profileParser = null;
+        }
     }
 
     /**
@@ -275,7 +296,7 @@ public final class RestrictedSecurity {
     private static void getProfileID(Properties props) {
         String potentialProfileID = "RestrictedSecurity." + selectedProfile;
 
-        if (selectedProfile.indexOf(".") != -1) {
+        if (selectedProfile.indexOf('.') != -1) {
             /* The default profile is used, or the user specified the
              * full <profile.version>.
              */
@@ -345,8 +366,8 @@ public final class RestrictedSecurity {
         }
     }
 
-    private static void checkFIPSCompatibility(Properties props) {
-        boolean isFIPSProfile = Boolean.parseBoolean(props.getProperty(profileID + ".desc.fips"));
+    private static void checkFIPSCompatibility() {
+        boolean isFIPSProfile = restricts.descIsFIPS;
         if (isFIPSProfile) {
             if (debug != null) {
                 debug.println("RestrictedSecurity profile " + profileID
@@ -428,13 +449,9 @@ public final class RestrictedSecurity {
                 getProfileID(props);
                 checkIfKnownProfileSupported();
 
-                // If user enabled FIPS, check whether chosen profile is applicable.
-                if (userEnabledFIPS) {
-                    checkFIPSCompatibility(props);
-                }
-
                 // Initialize restricted security properties from java.security file.
-                restricts = new RestrictedSecurityProperties(profileID, props);
+                profileParser = new ProfileParser(profileID, props);
+                restricts = profileParser.getProperties();
 
                 // Restricted security properties checks.
                 restrictsCheck();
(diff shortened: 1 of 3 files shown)


KostasTsiounis:
FYI @keithc-ca

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -610,31 +616,66 @@ private static void printStackTraceAndExit(String message) {
...
+        private final String jdkTlsEphemeralDHKeySize;
+        private final String jdkTlsLegacyAlgorithms;
+        private final String jdkCertpathDisabledAlgorithms;
+        private final String jdkSecurityLegacyAlgorithms;
+        private final String keyStoreType;
+        private final String keyStore;
 
         // For SecureRandom.
-        String jdkSecureRandomAlgorithm;
+        final String jdkSecureRandomProvider;
```
keithc-ca, line 675: This was not needed before: why is it added?

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -919,34 +752,52 @@ boolean isRestrictedServiceAllowed(Service service) {
                 return false;
             } else if (constraints.length == 0) {
                 // Allow this provider with no constraints.
+                if (debug != null) {
```
keithc-ca, line 755: Please copy the debug code from later versions, before line 752, and add a blank line before line 848.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1039,30 +900,605 @@ boolean isRestrictedProviderAllowed(String providerName) {
...
+            }
+
+            // If there's more, check that all of the subsequent ones are set to be removed.
+            for (int i = numOfExistingProviders; i >= providerPos; i--) {
+                if (debug != null) {
+                    debug.println("\t\t\tNot the last provider. More to be removed.");
+                }
+
+                String providerInfo = securityProps.getProperty(profileExtensionId + ".jce.provider." + i);
+                if ((providerInfo == null) || !providerInfo.isBlank()) {
```
keithc-ca, line 1215: `String.isBlank()` is only available in Java 11+. Suggest splitting `isNullOrBlank()`:
```java
    /**
     * Check if the input string is blank.
     *
     * @param string the input string
     * @return true if the input string is blank
     */
    private static boolean isBlank(String string) {
        return string.trim().isEmpty();
    }

    /**
     * Check if the input string is null or blank.
     *
     * @param string the input string
     * @return true if the input string is null or blank
     */
    private static boolean isNullOrBlank(String string) {
        return (string == null) || isBlank(string);
    }
```
so the local `isBlank()` can be used instead.

keithc-ca (changes requested):

jdk/src/share/lib/security/java.security-linux
```diff
@@ -138,6 +139,7 @@ RestrictedSecurity.NSS.140-2.jce.provider.4 = com.sun.net.ssl.internal.ssl.Provi
 RestrictedSecurity.NSS.140-2.keystore.type = PKCS11
 RestrictedSecurity.NSS.140-2.javax.net.ssl.keyStore = NONE
 
+RestrictedSecurity.NSS.140-2.securerandom.provider = SunPKCS11-NSS-FIPS
```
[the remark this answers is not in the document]
KostasTsiounis: It's additional information that is available in all versions, but not 8. It is not really used anywhere at the moment, but I figured it would be nice to be consistent.

I can remove it, if you think it shouldn't be there.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Same as above.

jdk/src/share/lib/security/java.security-linux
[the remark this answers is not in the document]
keithc-ca: I think it should be removed: leaving it here would suggest it's something a user could configure, but it isn't used. Worse, if omitted, the profile would be rejected.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -179,6 +184,22 @@ public static boolean isEnabled() {
...
+     *
+     * @return the secure random provider
+     */
+    public static String getRandomProvider() {
+        if (!securityEnabled) {
+            printStackTraceAndExit(
+                    "Restricted security mode secure random provider can only be used when restricted security mode is enabled.");
+        }
+        return restricts.jdkSecureRandomProvider;
+    }
```
keithc-ca, line 201: This method should be removed - it's not used.
The field `jdkSecureRandomProvider` and all references  should likewise be removed.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -1101,145 +940,844 @@ private void listUsedProfile() {
...
+                        // Append additional values to property.
+                        value = value.substring(1).trim();
+
+                        // Take existing value of property into account, if applicable.
+                        if (existingValue == null) {
+                            printStackTraceAndExit("Property '" + property + "' does not exist in parent profile. Cannot append.");
+                        } else if (isBlank(existingValue)) {
+                            newValue = value;
+                        } else {
+                            newValue = (isBlank(value)) ? existingValue : existingValue + ", " + value;
```
keithc-ca, line 1635: Please remove the redundant parentheses:
```java
                            newValue = isBlank(value) ? existingValue : existingValue + ", " + value;
```

keithc-ca (changes requested):

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed.

jdk/src/share/lib/security/java.security-linux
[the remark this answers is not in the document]
KostasTsiounis: Removed.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed.

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Done.

keithc-ca:
The unexpected copyright check failure should be fixed by https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/757.

keithc-ca:
Jenkins test sanity win jdk8

keithc-ca (approved):

PR #758, opened 2024-06-26
Support provider fully-qualified class name in Restricted Security mode

taoliult:
This is a back port PR from JDKNext PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/758

This PR is for updating the codes to only support the fully-qualified class name when checking the providers in restricted security mode. And because of the provider list in OpenJDK8 java.security file is already using the fully-qualified class name, then no changes needed for the java.security file.


keithc-ca:

keithc-ca (approved):

PR #761, opened 2024-08-30
Merge master jdk8u432-b03 into openj9-staging

JasonFengJ9:
Merge master jdk8u432-b03 into openj9-staging

Conflicts:
```
	jdk/THIRD_PARTY_README
```


keithc-ca:
The conflict should be resolved by adding the FreeType license to `THIRD_PARTY_README` as was done in `jdk/THIRD_PARTY_README` for #737.

JasonFengJ9:
> The conflict should be resolved by adding the FreeType license to THIRD_PARTY_README as was done in jdk/THIRD_PARTY_README for https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/737.

The OpenJDK update caused the conflict was [[8u] Remove duplicate license files](https://bugs.openjdk.org/browse/JDK-8338144)
_**Description**
As a legacy of the split repository structure 8u had in the past, it has duplicate copies of THIRD_PARTY_README, LICENSE and ASSEMBLY_EXCEPTION in both the root of the source tree and each of the former subrepository roots (hotspot, jdk, nashorn, jaxp, jaxws, corba, langtools)
Of these, LICENSE and ASSEMBLY_EXCEPTION do not tend to change and so are not that problematic, but THIRD_PARTY_README often has to be updated when third-party in-tree libraries are updated, requiring duplicate changes to every copy.
We should remove the duplicates and just have one copy in the top of the source tree. This is already true for LICENSE and ASSEMBLY_EXCEPTION in 11u and later, and **THIRD_PARTY_README has been replaced by individual licenses in Markdown format under the 'legal' subdirectories.**_

Any particular reason that FreeType can't have an individual license in its folder?

pshipton:
The original change was https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/632

We added freetype to jdk8 (it doesn't come from OpenJDK 8), and so the license needs to be included as well.
jdk8 doesn't have individual license folders, just jdk11+.

JasonFengJ9:
OpenJDK https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/761/commits/800a107c7871838bf093337ce99e9c6c0d93c03b removed `jdk/THIRD_PARTY_README`.
There is still https://github.com/ibmruntimes/openj9-openjdk-jdk8/blob/openj9/THIRD_PARTY_README
Maybe move the FreeType license there?

pshipton:
>Maybe move the FreeType license there?

Yes I agree. I assume that is now the file that ends up in the root directory of the JVM build.

keithc-ca (approved):

PR #763, opened 2024-09-18
Ensure static fields set during signed jar process

jasonkatonica:
When loading a signed jar file that is on the classpath, such as the
bouncy castle signed JCE jar file, it has been observed that the value
of `jarVerificationProviders` is set to null. This causes a
NullPointerException during the loading process.

This update moves the static declaration of `jarVerificationProviders`
to be prior to the method call `RestrictedSecurity.checkHashValues()`
since this method call needs the value of the field
`jarVerificationProviders` to be initialized in order to work correctly.

Signed-off-by: Jason Katonica <katonica@us.ibm.com>

### DIFF
--- a/jdk/src/share/classes/sun/security/jca/Providers.java
+++ b/jdk/src/share/classes/sun/security/jca/Providers.java
@@ -55,14 +55,6 @@ public class Providers {
     // Note volatile immutable object, so no synchronization needed.
     private static volatile ProviderList providerList;
 
-    static {
-        // set providerList to empty list first in case initialization somehow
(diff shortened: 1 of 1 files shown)


keithc-ca:
Please update the commit message and the description here (there's no `restrictedJarVerificationProviders` array in jdk8).

jasonkatonica:
> Please update the commit message and the description here (there's no `restrictedJarVerificationProviders` array in jdk8).

Updated the commit message and PR description message as suggested.

keithc-ca (approved):

PR #764, opened 2024-09-30
Remove redundant exports on Windows

keithc-ca:
`INCLUDE`, `LIB`, and `PATH` are already exported in spec.gmk.

See also https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/850 and back-ports.


pshipton:
jenkins compile win jdk8

pshipton (approved):

PR #767, opened 2024-10-04
Allow RestrictedSecurity property extension from default values

KostasTsiounis:
The default values existing in the java.security file are now considered the baseline and all profiles can inherit them and add or remove to them, if they are in list form. This does not include the providers, though.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/849


KostasTsiounis:
FYI @keithc-ca

keithc-ca:

PR #769, opened 2024-10-16
Add more granular digest flags

KostasTsiounis:
Instead of just having a single flag to enable/disable native support for all digest algorithms, separate flags are created for each one of them.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/862


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins test sanity aix jdk8

PR #771, opened 2024-10-28
Backport 8307850 to jdk8

pshipton:
8307850: update for deprecated sprintf for jdk.jdi
https://github.ibm.com/runtimes/openj9-openjdk-jdk17/commit/d1acd4d8943


jdk/src/share/transport/shmem/shmemBase.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2008, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2024, Oracle and/or its affiliates. All rights reserved.
```
keithc-ca, line 2: The second date is 2023 in jdk17; I suggest it should be the same here.

keithc-ca (changes requested):
Please update the commit message and the description to use the public URL for the referenced commit: https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/d1acd4d8943b910935523359ee6d9dd22a3a471b.

This should also include changes to `jdk/src/share/native/com/sun/tools/jdi/SharedMemoryTransport.c`.

jdk/src/share/transport/shmem/shmemBase.c
pshipton: Updated.

pshipton:
Updated.

keithc-ca:

keithc-ca (approved):

PR #772, opened 2024-10-28
Set AIX executables to have exempt SED status

zl-wang:
Customers complained IBM SDK8 has this status, but Semeru Runtimes don't, such that they 
cannot run Semeru Runtimes on their systems when migrating from IBM SDK8. This is a temporary 
fix without any other impacts, while we are investigating a complete solution for a larger issue 
(HugePages are not used for codeCache even though proper -Xlp option is specified).

### DIFF
--- a/common/autoconf/flags.m4
+++ b/common/autoconf/flags.m4
@@ -24,7 +24,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
 # ===========================================================================
 
 AC_DEFUN_ONCE([FLAGS_SETUP_INIT_FLAGS],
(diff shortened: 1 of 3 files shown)


common/autoconf/flags.m4
```diff
@@ -440,7 +440,7 @@ AC_DEFUN_ONCE([FLAGS_SETUP_COMPILER_FLAGS_FOR_JDK],
       CXXFLAGS_JDKLIB_EXTRA="${CXXFLAGS_JDKLIB_EXTRA} -xregs=no%appl"
     fi
   elif test "x$TOOLCHAIN_TYPE" = xxlc; then
-    LDFLAGS_JDK="${LDFLAGS_JDK} -q64 -brtl -bnolibpath -liconv -bexpall"
+    LDFLAGS_JDK="${LDFLAGS_JDK} -q64 -brtl -bnolibpath -liconv -bexpall -brwexec_must"
```
keithc-ca, line 443: This is incomplete: in jdk8, you must run `bash common/autoconf/autogen.sh` to update the `generated-configure.sh` files and add them to the commit.

keithc-ca (changes requested):

keithc-ca:
Please rebase to address the merge conflict.

keithc-ca:
Jenkins test sanity.functional aix,plinux jdk8

keithc-ca (approved):

PR #779, opened 2024-11-05
Check if OpenSSL is in FIPS mode and use Java for some algorithms

KostasTsiounis:
If the `OpenSSL` library used is in `FIPS` mode, avoid using it for algorithms that are not `FIPS` compliant and revert to the original Java implementation.

Back-ported by: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/867

### DIFF
--- a/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -86,6 +86,8 @@ public class NativeCrypto {
             return osslVersion;
         }).longValue();
 
+    private static final boolean isOpenSSLFIPS = (ossl_ver != -1) && isOpenSSLFIPS();
+
     /**
      * Return the OpenSSL version. It will return -1 if the library isn't
      * successfully loaded.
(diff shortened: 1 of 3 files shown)


KostasTsiounis:
FYI @keithc-ca

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -86,6 +86,17 @@ public class NativeCrypto {
...
 
+    private static final boolean isOpenSSLFIPS = AccessController.doPrivileged(
+        (PrivilegedAction<Boolean>) () -> {
+        boolean isFIPS;
+        if (ossl_ver != -1) {
+            isFIPS = isOpenSSLFIPS();
+        } else {
+            isFIPS = false;
+        }
+        return isFIPS;
```
keithc-ca, line 97: This doesn't require privileges; it can be simplified to:
```java
    private static final boolean isOpenSSLFIPS = (ossl_ver != -1) && isOpenSSLFIPS();
```

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -189,6 +200,52 @@ public static final boolean isTraceEnabled() {
     private NativeCrypto() {
         //empty
     }
+    public static final boolean isOpenSSLFIPSVersion() {
```
keithc-ca, line 203: Please insert a blank line before this.

keithc-ca (changes requested):

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Yes, you're right. Changed to suggestion.

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
keithc-ca: Actually, I suggest this method is not needed at all - it's only used in this class. Just use `isOpenSSLFIPS` directly.

keithc-ca (changes requested):

closed/adds/jdk/src/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
KostasTsiounis: Removed method altogether.

keithc-ca:
Jenkins compile zlinux jdk8

keithc-ca (approved):

PR #782, opened 2024-11-10
Add the error message when the profile does not exist

taoliult:
This is a back-port PR from PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/875

In Restricted Security Mode, add the error message when the profile name is used without the version part in "-Dsemeru.customprofile=", and the specified profile does not exist in the java.security file.


keithc-ca:
Jenkins compile zlinux jdk8

keithc-ca (approved):

PR #791, opened 2025-02-01
Remove --without-cmake support

pshipton:
Backport https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/928

### DIFF
--- a/closed/CopyToBuildJdk.gmk
+++ b/closed/CopyToBuildJdk.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2020, 2024 All Rights Reserved
+# (c) Copyright IBM Corp. 2020, 2025 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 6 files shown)


keithc-ca:
Perhaps we should keep UMA support at least for this version? Otherwise it will make it more challenging to keep it working for IBM Java 8 builds.

pshipton:
It seems easy enough, and better targeted, to just run IBM Java 8 builds for anything of concern.

keithc-ca:
Not as easy as doing a local build, but workable.

closed/CopyToBuildJdk.gmk
```diff
@@ -216,12 +216,7 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
   $(OPENJ9_VM_BUILD_DIR)/j9ddr.dat : run-ddrgen
 
   run-ddrgen :
-    ifeq (true,$(OPENJ9_ENABLE_CMAKE))
 	  $(EXPORT_MSVS_ENV_VARS) $(MAKE) -C $(OPENJ9_VM_BUILD_DIR) j9ddr
```
keithc-ca, line 219: Could you please trim the leading spaces from this line (that wouldn't be here if it were never conditional)?

jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2024 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2025 All Rights Reserved
```
keithc-ca, line 2: Changes to this file require running `bash common/autoconf/autogen.sh` to update `generated-configure.sh` files.

keithc-ca (changes requested):

jdk/make/closed/autoconf/custom-hook.m4
pshipton: Updated the files and removed the spaces in CopyToBuildJdk.gmk

keithc-ca:
Jenkins compile zlinux jdk8

keithc-ca (approved):

PR #794, opened 2025-02-20
Backport 8347740: java/io/File/createTempFile/SpecialTempFile.java failing

pshipton:
Fix the test so it doesn't fail on Windows 11 or Server 2025 machines.

https://bugs.openjdk.org/browse/JDK-8345368
https://bugs.openjdk.org/browse/JDK-8347740

https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/0ff2c3742e

### DIFF
--- a/jdk/test/java/io/File/createTempFile/SpecialTempFile.java
+++ b/jdk/test/java/io/File/createTempFile/SpecialTempFile.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


jdk/test/java/io/File/createTempFile/SpecialTempFile.java
```diff
@@ -21,6 +21,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * ===========================================================================
+ */
+
```
keithc-ca, line 29: I don't think we want to add this, or did I miss some changes that weren't just back-ported?

keithc-ca (changes requested):

jdk/test/java/io/File/createTempFile/SpecialTempFile.java
pshipton: I had to modify `OSVersion.current()` to `System.getProperty("os.version")`

keithc-ca: Ok, but then it should say 2025, not 2022.

pshipton: Ya, updated.

keithc-ca (approved):

PR #799, opened 2025-03-06
Allow multiple constraints for each algorithm

KostasTsiounis:
If a constraint for an algorithm is found and the class attempting to utilize it doesn't match the accepted uses, the algorithm is considered not allowed and loading it does not succeed.

Instead, we want to check all available constraints for an algorithm before deciding if it is allowed to be used by a specific module, package and/or class.

Additional test cases are added to check this functionality.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/954

### DIFF
--- a/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
+++ b/closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
@@ -72,6 +72,16 @@ public class TestConstraintsSuccess {
         KeyManagerFactory.getInstance("SunX509");
         TrustManagerFactory.getInstance("SunX509");
         SSLContext.getInstance("TLSv1.3");
+
+        // Since there are three constraints for MD5, with only the middle one
+        // allowing for use by this class, successfully getting the algorithm
+        // verifies that all constraints are checked.
+        MessageDigest.getInstance("MD5");
+
+        // Since there are three constraints for SHA1withECDSA, with only the
+        // middle one having the correct attributes, successfully getting the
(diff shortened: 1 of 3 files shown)


closed/test/jdk/openj9/internal/security/TestConstraintsSuccess.java
```diff
@@ -72,6 +72,16 @@ private static void getInstances() throws Exception {
...
+
+        // Since there are three constraints for MD5, with only the middle one
+        // allowing for use by this class, successfully getting the algorithm
+        // verifies that all constraints are checked.
+        MessageDigest.getInstance("MD5");
+
+        // Since there are three constraints for SHA1withECDSA, with only the
+        // middle one having the correct attributes, successfully getting the
+        // algorithm verifies that all constraints are checked.
+        Signature.getInstance("SHA1withECDSA");
```
keithc-ca, line 84: Why does this not use `SHA512withECDSA` like other Java versions?

KostasTsiounis: `SHA512withECDSA` in Java 8 doesn't have `KeySize` in the attributes, so that I could have multiple constraints with different sizes. `SHA1withECDSA` does, however, allowing me to have one before and one after with the wrong key size.

keithc-ca: That makes sense. Thanks.

keithc-ca:

keithc-ca (approved):

PR #800, opened 2025-03-06
Set property com.ibm.fips.mode based on active FIPS profile

KostasTsiounis:
When loading a `RestrictedSecurity` `FIPS` profile, we need to set the system property value `com.ibm.fips.mode` to the value contained within the active profile.

A `RestrictedSecurity` profile, however, doesn't have to be a `FIPS` profile, in which case a `FIPS mode` property is not required and the corresponding system property should not be set.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/957 and https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/756


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #803, opened 2025-03-20
Use proper library path based on architecture

KostasTsiounis:
In Java8 there is another level of directories under the `lib` directory pertaining to the specific architecture for which the JDK is built.

To address that difference, separate specification of path per architecture is required.

### DIFF
--- a/closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
+++ b/closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
@@ -646,7 +646,7 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
 #if defined(_WIN32)
         static const char pathSuffix[] = "\\bin\\";
 #else /* defined(_WIN32) */
-        static const char pathSuffix[] = "/lib/";
+        static const char pathSuffix[] = "/lib" OPENJDK_TARGET_CPU_LIBDIR "/";
 #endif /* defined(_WIN32) */
 
         size_t path_len = strlen(chomepath) + sizeof(pathSuffix) - 1;
--- a/jdk/make/lib/SecurityLibraries.gmk
+++ b/jdk/make/lib/SecurityLibraries.gmk
@@ -319,6 +319,7 @@ ifeq ($(WITH_OPENSSL), yes)
         LANG := C, \
         OPTIMIZATION := LOW, \
         CFLAGS := $(CFLAGS_JDKLIB) \
+            -DOPENJDK_TARGET_CPU_LIBDIR='"$(OPENJDK_TARGET_CPU_LIBDIR)"' \
             -I$(SRC_ROOT)/closed/adds/jdk/src/share/native/jdk/crypto/jniprovider \
             -c $(OPENSSL_CFLAGS), \
         MAPFILE := $(SRC_ROOT)/closed/make/mapfiles/libjncrypto/mapfile-vers, \


closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
```diff
@@ -643,11 +643,25 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
 
     /* If JAVA_HOME is not null or empty and no library has been loaded yet, try there. */
     if ((NULL != chomepath) && ('\0' != *chomepath) && (NULL == crypto_library)) {
-#if defined(_WIN32)
-        static const char pathSuffix[] = "\\bin\\";
-#else /* defined(_WIN32) */
```
keithc-ca, line 648: I think Windows should continue to be checked first (if ever support for Windows on aarch64 were added, this would yield the wrong path). This should result in the same path as captured in `LIBCRYPTO_TARGET_LIB` in `jdk/make/CopyFiles.gmk`.

keithc-ca (changes requested):

closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
keithc-ca: An alternative, simpler strategy would be to change `CopyFiles.gmk` to put the shared library directly in the `lib` directory like the code currently expects.

KostasTsiounis: I could move it first or change the first to `#if defined(__aarch64__) && defined(__linux__)` to be more strict. Which one do you think is preferable? 

We could move it to `lib` but, given that all of the dll libraries on Windows are under `bin`, I would rather not have `OpenSSL` be the exception.

keithc-ca: I wasn't suggesting a change for Windows, just the other platforms where it is currently in `lib/<arch>` it would go in `lib`.

pshipton: I'm not sure we should do that. There aren't any other .so files in the lib directory. The JVM sets up some paths which probably don't include lib.

KostasTsiounis: Moved Windows and macOS first.

closed/adds/jdk/src/share/native/jdk/crypto/jniprovider/NativeCrypto.c
```diff
@@ -643,11 +643,21 @@ find_crypto_library(jboolean traceEnabled, const char *chomepath)
...
+        static const char pathSuffix[] = "/lib/aarch64/";
+#elif defined(__ppc64__) && defined(_AIX) /* defined(__aarch64__) */
+        static const char pathSuffix[] = "/lib/ppc64/";
+#elif defined(__PPC64__) && defined(__linux__) /* defined(__ppc64__) && defined(_AIX) */
+        static const char pathSuffix[] = "/lib/ppc64le/";
+#elif defined(__s390x__) /* defined(__PPC64__) && defined(__linux__) */
+        static const char pathSuffix[] = "/lib/s390x/";
+#else /* defined(__s390x__) */
+        static const char pathSuffix[] = "/lib/amd64/";
+#endif /* defined(__APPLE__) */
```
keithc-ca, line 660: This should use `OPENJDK_TARGET_CPU_LIBDIR` instead of duplicating the logic to determine the subdirectory.
```c
#if defined(_WIN32)
        static const char pathSuffix[] = "\\bin\\";
#else /* defined(_WIN32) */
        static const char pathSuffix[] = "/lib" OPENJDK_TARGET_CPU_LIBDIR "/";
#endif /* defined(_WIN32) */
```
This approach is enabled by adding to `CFLAGS` for `BUILD_JNCRYPTO` in `SecurityLibraries.gmk`:
```
            -DOPENJDK_TARGET_CPU_LIBDIR='"$(OPENJDK_TARGET_CPU_LIBDIR)"' \
```

keithc-ca (changes requested):

keithc-ca:
Please squash.

keithc-ca:
Jenkins test sanity plinux,win jdk8

keithc-ca:
Spot-checking the ppc64le build:
```
$ strings libjncrypto.so | grep /lib/
/lib/ppc64le/
```

keithc-ca (approved):

PR #804, opened 2025-03-21
Include System.c natives to make use of System.initProperties

pshipton:
Remove the Java_java_lang_System_mapLibraryName native so it can't conflict with the OpenJ9 native of the same name.

Related to https://github.com/eclipse-openj9/openj9/issues/21189


keithc-ca:

keithc-ca (approved):

PR #813, opened 2025-04-08
Set 32-bit OpenJ9 codes

pshipton:


pshipton:
@keithc-ca I suggest we want to include this in the 0.51 release. Not because of adding the 32-bit codes, but because `jdk/make/closed/autoconf/generated-configure.sh` wasn't regenerated for https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/797 and I believe Windows 32-bit is broken without the `-o "x$OPENJDK_TARGET_CPU_BITS" = x32` check.

pshipton:
Perhaps it's worth auto-generating the generated-configure.sh files on every build like jdk11+, not sure how big of a change that is.

keithc-ca:
Jenkins compile xlinux,win,win32,zlinux jdk8

keithc-ca (approved):
All builds have completed or are well past any effects of this change.

keithc-ca:
> I was thinking the same thing, but, like you, I don't know how much work is involved.

It wasn't as much work as I anticipated; see https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/814.

PR #816, opened 2025-04-17
Check RestrictedSecurity profile hash after Providers init

KostasTsiounis:
If the `Providers` class hasn't had a chance to initialize, the required message digest algorithms are not available to check the hash value of the `RestrictedSecurity` profile.

With this change, the check is only allowed after the appropriate initialization has been performed.

Issue https://github.com/eclipse-openj9/openj9/issues/21615

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/991


KostasTsiounis:
FYI @keithc-ca

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -80,6 +80,8 @@ public final class RestrictedSecurity {
 
     private static ProfileParser profileParser;
 
+    private static boolean enableCheckHashes;
```
keithc-ca, line 83: Please remove the unnecessary annotation on line 99.

keithc-ca (changes requested):

closed/adds/jdk/src/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed.

keithc-ca:

keithc-ca:
The test failure is known: https://github.com/eclipse-openj9/openj9/issues/9969.

keithc-ca (approved):

PR #822, opened 2025-05-05
Add support for AIX 32-bit builds

pshipton:
See also https://github.com/eclipse-openj9/openj9/pull/21801

### DIFF
--- a/common/autoconf/flags.m4
+++ b/common/autoconf/flags.m4
@@ -24,7 +24,7 @@
 #
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2025 All Rights Reserved
 # ===========================================================================
 
 AC_DEFUN_ONCE([FLAGS_SETUP_INIT_FLAGS],
@@ -41,7 +41,7 @@ AC_DEFUN_ONCE([FLAGS_SETUP_INIT_FLAGS],
   if test "x$OPENJDK_TARGET_OS" = xmacosx; then
     ARFLAGS="-r"
   elif test "x$OPENJDK_TARGET_OS" = xaix; then
-    ARFLAGS="-X64"
+    ARFLAGS="-X$OPENJDK_TARGET_CPU_BITS"
   elif test "x$OPENJDK_TARGET_OS" = xwindows; then
     # lib.exe is used as AR to create static libraries.
     ARFLAGS="-nologo -NODEFAULTLIB:MSVCRT"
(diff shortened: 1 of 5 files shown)


common/autoconf/flags.m4
```diff
@@ -41,7 +41,11 @@ AC_DEFUN_ONCE([FLAGS_SETUP_INIT_FLAGS],
   if test "x$OPENJDK_TARGET_OS" = xmacosx; then
     ARFLAGS="-r"
   elif test "x$OPENJDK_TARGET_OS" = xaix; then
-    ARFLAGS="-X64"
+    if test "x$OPENJDK_TARGET_CPU_BITS" = "x32"; then
+      ARFLAGS="-X32"
+    else
+      ARFLAGS="-X64"
+    fi
```
keithc-ca, line 48: Things like this need not be conditional:
```
    ARFLAGS="-X$OPENJDK_TARGET_CPU_BITS"
```
and lines 447-455 can be simplified likewise
```
    LDFLAGS_JDK="${LDFLAGS_JDK} -q$OPENJDK_TARGET_CPU_BITS -brtl -bnolibpath -liconv -bexpall -brwexec_must"
    CFLAGS_JDK="${CFLAGS_JDK} -qchars=signed -q$OPENJDK_TARGET_CPU_BITS -qfullpath -qsaveopt -qstackprotect"
    CXXFLAGS_JDK="${CXXFLAGS_JDK} -qchars=signed -q$OPENJDK_TARGET_CPU_BITS -qfullpath -qsaveopt -qstackprotect"
```

common/autoconf/platform.m4
```diff
@@ -256,11 +256,13 @@ AC_DEFUN([PLATFORM_SETUP_TARGET_CPU_BITS],
         OPENJDK_TARGET_CPU=x86
       elif test "x$OPENJDK_TARGET_CPU_ARCH" = "xsparc"; then
         OPENJDK_TARGET_CPU=sparc
+      elif test "x$OPENJDK_TARGET_CPU_ARCH" = "xppc"; then
+        OPENJDK_TARGET_CPU=ppc
```
keithc-ca, line 260: I think `closed/openssl.gmk` needs to be updated to handle `ppc`.

keithc-ca (changes requested):

common/autoconf/platform.m4
pshipton: There is no need of openssl for 32-bit support.

common/autoconf/flags.m4
pshipton: Updated.

common/autoconf/platform.m4
keithc-ca: `configure` should fail if openssl is enabled for unsupported 32-bit platforms rather than build a 64-bit library.

closed/openssl.gmk
```diff
@@ -63,10 +63,12 @@ endif # windows
...
-  ifeq ($(TOOLCHAIN_TYPE), clang)
-    OPENSSL_TARGET := aix64-clang
-  else ifeq ($(TOOLCHAIN_TYPE), xlc)
-    OPENSSL_TARGET := aix64-cc
+  ifeq ($(OPENJDK_TARGET_CPU), ppc64)
+    ifeq ($(TOOLCHAIN_TYPE), clang)
+      OPENSSL_TARGET := aix64-clang
+    else ifeq ($(TOOLCHAIN_TYPE), xlc)
+      OPENSSL_TARGET := aix64-cc
+    endif
```
keithc-ca, line 71: I was expecting the misconfiguration would be detected by `configure`, not later during a build (attempt).

keithc-ca: Worse, if openssl appears to be built (`libcrypto.so` and `libssl.so` are present), it doesn't try to build it, but would use (bundle) an incompatible version.

pshipton: Updated.

keithc-ca:
Jenkins compile aix,win32 jdk8

keithc-ca (approved):

PR #825, opened 2025-05-12
Use CXX_VERSION_STRING instead of COMPILER_VERSION_STRING

pshipton:
COMPILER_VERSION_STRING seems an internal variable that shouldn't be
used. Also remove the redundant init of COMPILER_VERSION_STRING since
CXX_VERSION_STRING is set in toolchain.m4.

Partial backport of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1010


jdk/make/closed/autoconf/custom-hook.m4
```diff
@@ -536,33 +536,6 @@ AC_DEFUN_ONCE([CUSTOM_LATE_HOOK],
...
-    #     Copyright (C) 2013 Free Software Foundation, Inc.
-    #     This is free software; see the source for copying conditions.  There is NO
-    #     warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
-    COMPILER_VERSION_OUTPUT=`$COMPILER --version 2>&1`
-    # Remove Copyright and legalese from version string, and
-    # collapse into a single line
-    COMPILER_VERSION_STRING=`$ECHO $COMPILER_VERSION_OUTPUT | \
-        $SED -e 's/ *Copyright .*//'`
-  fi
   AC_SUBST(COMPILER_VERSION_STRING)
```
keithc-ca, line 539: There are two compiler version strings (`CC_VERSION_STRING` and `CXX_VERSION_STRING`) we should use one (or both) of those directly instead of `COMPILER_VERSION_STRING` which looks otherwise like a local variable in `TOOLCHAIN_EXTRACT_COMPILER_VERSION`.
```
  AC_SUBST(CXX_VERSION_STRING)
```
`custom-spec.gmk.in` and `openj9_version_info.h.in` should be updated to use this name (and `CC_VERSION_STRING` if necessary).

keithc-ca (changes requested):

jdk/make/closed/autoconf/custom-hook.m4
pshipton: Pls review https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1010

pshipton: Updated with the backported changes.

keithc-ca:
Please rebase to avoid an implicit merge with https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/824.

pshipton:
Rebased.

keithc-ca:
Jenkins compile win32 jdk8

keithc-ca (approved):

PR #847, opened 2025-09-04
Support a list of java.security property files

JinhangZhang:
This PR proposes a new system property named `java.security.propertiesList`. This new property will allow multiple java.security profiles from different files to be used rather than put those different profiles into a single java.security file. This new property supports a way to configure the superset of all files in this set of property files.


jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,83 +124,133 @@ private static void initialize() {
...
-                        propURL = new URL(extraPropFile);
-                    }
-                    bis = new BufferedInputStream(propURL.openStream());
-                    props.load(bis);
-                    loadedProps = true;
+                        file = PropertyExpander.expand(file);
+                        propFile = new File(file);
+                        if (propFile.exists()) {
+                            propURL = new URL
+                                    ("file:" + propFile.getCanonicalPath());
```
keithc-ca, line 143: Suggest this line need not be split:
```java
                            propURL = new URL("file:" + propFile.getCanonicalPath());
```

jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,83 +124,133 @@ private static void initialize() {
...
+                        props.load(bis);
+                        loadedProps = true;
 
-                    if (sdebug != null) {
-                        sdebug.println("reading security properties file: " +
-                                        propURL);
-                        if (overrideAll) {
+                        if (sdebug != null) {
+                            sdebug.println("reading security properties file: " +
+                                            propURL);
```
keithc-ca, line 153: Please don't split, use `file` like line 159.
```java
                            sdebug.println("reading security properties file: " + file);
```

jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,83 +124,133 @@ private static void initialize() {
...
+                        }
+                    } catch (Exception e) {
+                        if (sdebug != null) {
                             sdebug.println
-                                ("overriding other security properties files!");
+                                    ("unable to load security properties from " +
+                                    file);
```
keithc-ca, line 159: Please don't split:
```java
                            sdebug.println("unable to load security properties from " + file);
```

jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,83 +124,133 @@ private static void initialize() {
...
+                            e.printStackTrace();
+                            loadedProps = false;
+                            break;
```
keithc-ca, line 162: This should not be only when debugging:
```java
                        }
                        loadedProps = false;
                        break;
                    } finally {
```

jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,83 +124,133 @@ private static void initialize() {
...
                 }
             }
-        }
+            if (!loadedProps) {
+                initializeStatic();
+                if (sdebug != null) {
+                    sdebug.println("unable to load security properties " +
+                            "-- using defaults");
+                }
+            }
```
keithc-ca, line 189: Please don't delete this line.

keithc-ca (changes requested):

jdk/src/share/classes/java/security/Security.java
JinhangZhang: move it to one line

jdk/src/share/classes/java/security/Security.java
JinhangZhang: updated

jdk/src/share/classes/java/security/Security.java
JinhangZhang: updated

jdk/src/share/classes/java/security/Security.java
```diff
@@ -124,82 +124,130 @@ private static void initialize() {
...
+                            sdebug.println("reading security properties file: " + file);
+                        }
+                    } catch (Exception e) {
+                        if (sdebug != null) {
                             sdebug.println
-                                ("overriding other security properties files!");
+                                    ("unable to load security properties from " + file);
+                            e.printStackTrace();
+                        }
+                        loadedProps = false;
```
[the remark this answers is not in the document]
JinhangZhang: removed

keithc-ca:
Jenkins test sanity plinux jdk8

PR #885, opened 2026-05-11
Run java/rmi/reliability/benchmark/bench/serial/Main with -Xmso512k

pshipton:
Increase the native stack size for x86 mac, and some other platforms. Some platforms already use 512k as the default. No platform has a default which is bigger than 512k atm.

Related to https://github.com/eclipse-openj9/openj9/issues/23666#issuecomment-4409738921

Closes https://github.com/eclipse-openj9/openj9/issues/23666

Passing grinder
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59944/

### DIFF
--- a/jdk/test/java/rmi/reliability/benchmark/bench/serial/Main.java
+++ b/jdk/test/java/rmi/reliability/benchmark/bench/serial/Main.java
@@ -22,6 +22,12 @@
  */
 
 /*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


keithc-ca (approved):

keithc-ca:
In later versions, this test is run with `-Xss2m`.

